PR #7, opened 2024-07-17
Optional Quantization

kozlov721:
- Offline graph preparation is performed even with disabled quantization
- Added tests for rvc3 and rvc4 without quantization

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -153,7 +153,7 @@ def prepare(
         converted_model_path = (
             converted_model_path_prefix / f"{model_name}.dlc"
         )
-    elif service in ["rvc2", "rvc3"]:
+    elif service in ["rvc2", "rvc3", "rvc3_non_quant"]:
         converted_model_path = (
             converted_model_path_prefix
             / "intermediate_outputs"
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -27,6 +27,7 @@ MODEL_CONFIGS = chain(
                 MNISTMetric,
                 ["input"],
                 "onnx",
+                "",
             ),
             (
                 service,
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -59,14 +63,26 @@
...
+                "onnx",
+                "rvc4.disable_calibration true rvc3.disable_calibration true",
             ),
         ]
         for service in [
             "rvc2",
             "rvc2_superblob",
             "rvc3",
             "rvc3_quant",
+            "rvc3_non_quant",
```
tersekmatija, line 83: Is "rvc3" quant or non-quant?

tests/conftest.py
```diff
@@ -159,9 +189,9 @@ def prepare(
     else:
         file_url = f"s3://luxonis-test-bucket/modelconverter-test/archives/{model_name}.tar.xz"
 
-    if "quant" not in service:
+    if "quant" not in service or "non_quant" in service:
```
tersekmatija, line 192: Can we make this more clear by just using "quant" when we want the quantize flow?

If in some case quant is only option, then we would just defined service with "_quant" in it?

tersekmatija (approved):

tests/conftest.py
kozlov721: "rvc3" is normal conversion but tested on the non-quant model. "rvc3_non_quant" is where calibration is explicitly disabled.

tests/conftest.py
kozlov721: Yeah, the tests would use some general refactoring eventually, they're quite convoluted. I'd probably do it all in some separate PR.

PR #10, opened 2024-07-18
Output Directory Name in CLI

kozlov721:
Moved `output_dir_name` from config to CLI arguments. Can be now specified for both `convert` and `infer` commands using `--output-dir` (or just `-o`) argument. For `modelconverter infer` it replaces the `--dest` argument.


conorsim (approved):
LGTM

PR #20, opened 2024-07-25
CLI Archive Command

kozlov721:
Added an `archive` command to the CLI. This command takes a model file and creates an NN Archive. 
This command runs outside of the docker container.

**Currently supported model formats:**
  - ONNX
  - tflite
  - IR
  - dlc (if `snpe-dlc-info` is in path)
  - har (if `hailo_sdk_client` is installed in the python environment) 

**Usage:**

```bash
modelconverter archive path/to/model_file --save-path gs://remote/path --put-file-plugin plugin 
```

### DIFF
--- a/modelconverter/__main__.py
+++ b/modelconverter/__main__.py
@@ -535,8 +535,17 @@ def archive(
     model_path = resolve_path(path, MODELS_DIR)
     cfg = archive_from_model(model_path)
     save_path = save_path or f"{cfg.model.metadata.name}.tar.xz"
-    if not save_path.endswith(".tar.xz"):
-        save_path += f"/{cfg.model.metadata.name}.tar.xz"
+    if save_path.endswith("tar.xz"):
+        compression = "xz"
+    elif save_path.endswith("tar.gz"):
+        compression = "gz"
+    elif save_path.endswith("tar.bz2"):
+        compression = "bz2"
+    else:
+        compression = "xz"
+
+    if not save_path.endswith(f".tar.{compression}"):
+        save_path += f"/{cfg.model.metadata.name}.tar.{compression}"
     archive_name = save_path.split("/")[-1]
     protocol = LuxonisFileSystem.get_protocol(save_path)
     if protocol != "file":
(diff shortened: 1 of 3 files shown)


modelconverter/__main__.py
```diff
@@ -535,8 +535,17 @@ def archive(
...
+        compression = "xz"
+    elif save_path.endswith("tar.gz"):
+        compression = "gz"
+    elif save_path.endswith("tar.bz2"):
+        compression = "bz2"
+    else:
+        compression = "xz"
+
+    if not save_path.endswith(f".tar.{compression}"):
+        save_path += f"/{cfg.model.metadata.name}.tar.{compression}"
```
tersekmatija, line 548: Could it happen that it's already named `{cfg.model.metadata.name}` but is just missing `.tar.{compression}`?

tersekmatija (approved):

modelconverter/__main__.py
kozlov721: That's possible, though I'm not sure what would be the right approach here. Assume the user meant to write `archive_name.tar.xz` or they actually did want to save it to `archive_name/archive_name.tar.xz`? I'd vouch for the second, I can imagine you could want to have a directory `resnet18` and inside be saving `resnet18.dlc`, `resnet18.blob` and `resnet18.tar.xz`.

tersekmatija: That makes sense, yes.

modelconverter/utils/nn_archive.py
```diff
@@ -218,3 +218,49 @@ def modelconverter_config_to_nn(
...
+        },
+    }
+
+    for name, shape in metadata.input_shapes.items():
+        archive_cfg["model"]["inputs"].append(
+            {
+                "name": name,
+                "shape": shape,
+                "layout": make_default_layout(shape),
+                "dtype": "float32",
```
conorsim, line 245: Couldn't it be possible to have int8 with dlc? Maybe also tflite? Would there be some way to detect that?

modelconverter/utils/nn_archive.py
```diff
@@ -218,3 +218,49 @@ def modelconverter_config_to_nn(
...
+                "input_type": "image",
+                "preprocessing": {
+                    "mean": None,
+                    "scale": None,
+                    "reverse_channels": False,
+                    "interleaved_to_planar": False,
```
conorsim, line 251: Is there some way to detect these? Is it safe to assume these "defaults"?

PR #23, opened 2024-08-05
TFLite Support for RVC2

kozlov721:
Added support for conversion from `tflite` files for RVC2 platform.

### DIFF
--- a/docker/rvc2/Dockerfile
+++ b/docker/rvc2/Dockerfile
@@ -13,14 +13,14 @@ COPY --link --from=BASE /usr/lib /usr/lib
 COPY --link --from=BASE /usr/lib_openvino /usr/lib
 COPY --link --from=BASE /usr/local/lib_38 /usr/local/lib
 
-RUN pip install pip-autoremove bsdiff4 && \
-    pip-autoremove -y pot addict pip-autoremove && \
-    pip cache purge
+COPY --link docker/rvc2/requirements.txt .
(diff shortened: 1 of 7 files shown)


klemen1999 (approved):
Are models exported with OpenVino>=2024.2.0 runnable on DepthAI since there the default version is lower (2022.1 based on [online blobconverter](https://blobconverter.luxonis.com/))

Otherwise LGTM

kozlov721:
>Are models exported with OpenVino>=2024.2.0 runnable on DepthAI

I tested the exported models with `modelconverter benchmark` and they do run fine, so I hope it won't cause any issues. I'll tag the latest RVC2 image before this change so we can quickly test with older version in case there will be an issue with RVC2 models not running on DepthAI.

tersekmatija:
> Are models exported with OpenVino>=2024.2.0 runnable on DepthAI

IMO this is highly dangerous. We'd rather test this on the whole ZOO as the MYRIAD was deprecated after 2022.3

tersekmatija:
Possibly the right path instead would be TFLITE -> ONNX -> IR?

I am mainly worried about "compile_tool" failing if IR includes any new opsets for specific operations. I'd imagine at least some were introduced and some updated? These are the models we should be testing.

kozlov721:
Another option might be to use the old versions when converting from ONNX and only switch to newer version of `mo` when converting from TFLITE. And print some warnings that this is not guaranteed to work.  Additionally, we might add a switch that would keep the old version even for TFLITE and try TFLITE -> ONNX conversion first.

PR #26, opened 2024-08-07
TFLITE Conversion Support for RVC2 and RVC3

kozlov721:
Added support for conversion from `tflite` format to RVC2 and RVC3 platforms.


klemen1999 (approved):
The export flow for tflite models is now`tflite -> onnx -> IR -> Blob` correct?
LGTM

PR #31, opened 2024-08-28
Layout Fixes

kozlov721:
- Fixed ONNX modification for layouts other than NCHW
- Added the `--layout` parameter to `snpe-onnx-to-dlc`


klemen1999 (approved):

klemen1999:
Some tests are failing though, we should look into that

PR #32, opened 2024-08-29
Updated Libraries and Dockerfiles

kozlov721:
- Updated code for use with newest `luxonis-ml`
- Removed `Dockerfile.public` images for RVC2, RVC3, and RVC4
- Removed `Dockerfile.base`
- Reduced the size of the RVC4 image by roughly 1GB


shared_with_container/configs/resnet18_IR.yaml
```diff
@@ -4,14 +4,12 @@ input_model: "s3://luxonis-test-bucket/modelconverter-test/resnet18-simplified.b
 
 mean_values: "imagenet"
 scale_values: "imagenet"
-reverse_input_channels: true
```
klemen1999, line 7: Is this becasue `reverse_input_channels` is deprecated?

kozlov721: Yeah, reversing channels is the default behavior for image inputs now and it'd be better to not have deprecated fields in the example configs.

klemen1999: Makes sense, in that case might be good to remove it on other configs as well: [multistage.yaml](https://github.com/luxonis/modelconverter/blob/96e0c0e1d72bd03a48192b70d0b1f8b680e3e729/shared_with_container/configs/multistage.yaml#L8) and in [defaults.yaml](https://github.com/luxonis/modelconverter/blob/96e0c0e1d72bd03a48192b70d0b1f8b680e3e729/shared_with_container/configs/defaults.yaml#L97). Also should we change the data_types [here](https://github.com/luxonis/modelconverter/blob/96e0c0e1d72bd03a48192b70d0b1f8b680e3e729/shared_with_container/configs/defaults.yaml#L103) to lowercase?

klemen1999 (approved):

PR #46, opened 2024-10-31
Fix on RVC2 benchmark script

ptoupas:
Fixed RVC2 benchmark script to work with models that have >1 inputs.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,3 @@
-default_language_version:
-    python: python3
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.1.2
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -1,3 +1,5 @@
+default_language_version:
```
klemen1999: We are not including this part in this (or other libraries) for now as I believe it is a bit setup specific. Can you remove it from the PR and have this change only locally?

klemen1999 (approved):
LGTM!

.pre-commit-config.yaml
ptoupas: Done

kozlov721 (approved):
LGTM

PR #48, opened 2024-11-08
Added dai_type, updated encoding flags, and removed reverse_input_channels flag 

ptoupas:
Main changes:
-  Updated the `encoding.from` and `encoding.to` flags to replace the `reverse_input_channels`.
-  Removed `reverse_input_channels`.
-  Added `dai_type` (its still not 100% compatible with all DAI data types).
-  Fixed a bug with mean and scale values on RVC2 conversion with multiple inputs.
-  Removed the extra nodes added to the onnx model when `mean=[0,0,0]` and `scale=[1,1,1]`
-  Added extra unit tests for the encoding

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,3 +1,5 @@
+default_language_version:
+    python: python3.11
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.1.2
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,3 @@
-default_language_version:
-    python: python3.11
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.1.2
--- a/README.md
+++ b/README.md
@@ -135,25 +135,25 @@ The `encoding` flag in the YAML configuration file allows you to specify color e
   ```
   This configuration specifies that the input data is in RGB format and will be converted to BGR format during processing.
 
-> \[!NOTE\] Note on Default Encoding
+> [!NOTE]
 > If the encoding is not specified in the YAML configuration, the default values are set to `encoding.from=RGB` and `encoding.to=BGR`.
 
-> \[!NOTE\] Note on Global Settings
+> [!NOTE]
 > Certain options can be set **globally**, applying to all inputs of the model, or **per input**. If specified per input, these settings will override the global configuration for that input alone. The options that support this flexibility include `scale_values`, `mean_values`, `encoding`, `data_type`, `shape`, and `layout`.
 
 #### NN Archive Configuration File
 
 In the NN Archive configuration, there are two flags related to color encoding control:
 
-- **`reverse_channels` (Deprecated)**:
-  Determines the input color format of the model: when set to *True*, the input is considered to be *"RGB"*, and when set to *False*, it is treated as *"BGR"*. This flag is deprecated and will be replaced by the `dai_type` flag in future versions.
 - **`dai_type`**:
   Provides a more comprehensive control over the input type compatible with the DAI backend. It is read by DepthAI to automatically configure the processing pipeline, including any necessary modifications to the input image format.
+- **`reverse_channels` (Deprecated)**:
+  Determines the input color format of the model: when set to *True*, the input is considered to be *"RGB"*, and when set to *False*, it is treated as *"BGR"*. This flag is deprecated and will be replaced by the `dai_type` flag in future versions.
 
-> \[!NOTE\] Note on Default Settings
-> If neither `dai_type` nor `reverse_channels` the input to the model is considered to be *"BGR"*.
+> [!NOTE]
+> If neither `dai_type` nor `reverse_channels` the input to the model is considered to be *"RGB"*.
 
-> \[!NOTE\] Note on Conflicting Settings
+> [!NOTE]
 > If both `dai_type` and `reverse_channels` are provided, the converter will give priority to `dai_type`.
 
 ### Sharing Files
@@ -264,7 +264,7 @@ modelconverter convert rvc2 input_model models/yolov6n.onnx \
                         outputs.2.name out_2
 ```
 
-> \[!WARNING\]
+> [!WARNING]
 > If you modify the default stages names (`stages.stage_name`) in the configuration file (`config.yaml`), you need to provide the full path to each stage in the command-line arguments. For instance, if a stage name is changed to `stage1`, use `stages.stage1.inputs.0.name` instead of `inputs.0.name`.
 
 ## Multi-Stage Conversion
--- a/README.md
+++ b/README.md
@@ -158,6 +158,7 @@ You can run the built image either manually using the `docker run` command or us
 1. Execute the conversion:
 
 - If using the `docker run` command:
+
   ```bash
   docker run --rm -it \
     -v $(pwd)/shared_with_container:/app/shared_with_container/ \
@@ -168,11 +169,15 @@ You can run the built image either manually using the `docker run` command or us
     convert <target> \
     --path <s3_url_or_path> [ config overrides ]
   ```
+
 - If using the `modelconverter` CLI:
+
   ```bash
   modelconverter convert <target> --path <s3_url_or_path> [ config overrides ]
   ```
+
 - If using `docker-compose`:
+
   ```bash
   docker compose run <target> convert <target> ...
   ```
@@ -207,6 +212,8 @@ modelconverter convert rvc2 input_model models/yolov6n.onnx \
                         outputs.1.name out_1 \
                         outputs.2.name out_2
 ```
+
+:warning: **Important Note**: If you modify the default stages names (`stages.stage_name`) in the configuration file (`config.yaml`), you need to provide the full path to each stage in the command-line arguments. For instance, if a stage name is changed to `stage1`, use `stages.stage1.inputs.0.name` instead of `inputs.0.name`.
 
 ## Multi-Stage Conversion
 
--- a/README.md
+++ b/README.md
@@ -156,6 +156,9 @@ In the NN Archive configuration, there are two flags related to color encoding c
 > [!NOTE]
 > If both `dai_type` and `reverse_channels` are provided, the converter will give priority to `dai_type`.
 
+> [!IMPORTANT]
+> Provide mean/scale values in the original color format used during model training (e.g., RGB or BGR). Any necessary channel permutation is handled internally—do not reorder values manually.
+
 ### Sharing Files
 
 When using the supplied `docker-compose.yaml`, the `shared_with_container` directory facilitates file sharing between the host and container. This directory is mounted as `/app/shared_with_container/` inside the container. You can place your models, calibration data, and config files here. The directory structure is:
--- a/README.md
+++ b/README.md
@@ -101,9 +101,15 @@ To enable GPU acceleration for `hailo` conversion, install the [Nvidia Container
 
 ## Running ModelConverter
 
-Configuration for the conversion predominantly relies on a `yaml` config file. For reference, see [defaults.yaml](shared_with_container/configs/defaults.yaml) and other examples located in the [shared_with_container/configs](shared_with_container/configs) directory.
+There are two main ways to execute configure the conversion process:
 
-However, you have the flexibility to modify specific settings without altering the config file itself. This is done using command line arguments. You provide the arguments in the form of `key value` pairs. For better understanding, see [Examples](#examples).
+1. **YAML Config File (Primary Method)**:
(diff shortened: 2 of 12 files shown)


.pre-commit-config.yaml
```diff
@@ -1,3 +1,5 @@
+default_language_version:
+    python: python3.11
```
kozlov721: This should be 3.8

README.md
```diff
@@ -101,9 +101,15 @@ To enable GPU acceleration for `hailo` conversion, install the [Nvidia Container
 
 ## Running ModelConverter
 
-Configuration for the conversion predominantly relies on a `yaml` config file. For reference, see [defaults.yaml](shared_with_container/configs/defaults.yaml) and other examples located in the [shared_with_container/configs](shared_with_container/configs) directory.
+There are two main ways to execute configure the conversion process:
 
-However, you have the flexibility to modify specific settings without altering the config file itself. This is done using command line arguments. You provide the arguments in the form of `key value` pairs. For better understanding, see [Examples](#examples).
+1. **YAML Config File (Primary Method)**:
+   The primary way to configure the conversion is through a YAML configuration file. For reference, you can check [defaults.yaml](shared_with_container/configs/defaults.yaml) and other examples located in the [shared_with_container/configs](shared_with_container/configs) directory.
+1. **NN Archive**:
```
kozlov721, line 113: Should NN Archive be the primary method instead? @klemen1999

README.md
```diff
@@ -200,13 +211,16 @@ Specify all options via the command line without a config file:
...
-                        reverse_input_channels True \
+                        inputs.0.encoding.from RGB \
+                        inputs.0.encoding.to BGR \
                         shape "[1,3,256,256]" \
                         outputs.0.name out_0 \
                         outputs.1.name out_1 \
                         outputs.2.name out_2
 ```
 
+:warning: **Important Note**: If you modify the default stages names (`stages.stage_name`) in the configuration file (`config.yaml`), you need to provide the full path to each stage in the command-line arguments. For instance, if a stage name is changed to `stage1`, use `stages.stage1.inputs.0.name` instead of `inputs.0.name`.
```
kozlov721: You can use the following snippet instead; it renders a bit nicer. 

```md
> [!WARNING]
> If you modify the default stages names (`stages.stage_name`) in the configuration file (`config.yaml`), you need to provide the full path to each stage in the command-line arguments. For instance, if a stage name is changed to `stage1`, use `stages.stage1.inputs.0.name` instead of `inputs.0.name`.
```

> [!WARNING]
> If you modify the default stages names (`stages.stage_name`) in the configuration file (`config.yaml`), you need to provide the full path to each stage in the command-line arguments. For instance, if a stage name is changed to `stage1`, use `stages.stage1.inputs.0.name` instead of `inputs.0.name`.

modelconverter/utils/config.py
```diff
@@ -134,41 +134,9 @@ class InputConfig(OutputConfig):
     ] = RandomCalibrationConfig()
     scale_values: Optional[Annotated[List[float], Field(min_length=1)]] = None
     mean_values: Optional[Annotated[List[float], Field(min_length=1)]] = None
-    reverse_input_channels: bool = False
```
kozlov721, line 137: IMHO we should keep `reverse_input_channels` as a property of the `InputConfig` so the rest of the code doesn't have to be changed much. Checking `if inp.reverse_input_channels` is much nicer then always doing `if inp.encoding.to != inp.encoding.from_`.

kozlov721 (changes requested):
In general LGTM, just left a comment regarding removing the `reverse_input_channels`. I'd propose keeping it as a property of `InputConfig`, somehting like:

```python

@property
def reverse_input_channels(self) -> bool:
    return self.encoding.to != self.encoding.from_
```

That way we don't need to change the rest of the code where `inp.reverse_input_channels` is used. And it also looks much more readable than doing `inp.encoding.to != inp.encoding.from_` everywhere.

modelconverter/utils/config.py
klemen1999: If you mean keeping `reverse_input_channels` as a flag in the config.yaml I don't agree with this. It just polutes the namescape and `encoding` block is much clearer IMO. But if you mean having it just as an internal property of the `InputConfig` then this sounds good. But in that case I would rename it to something like `same_input_output(self)->bool` (or maybe better name) which would just compare `to` and `from` encodings and check if there are the same. I would just stir away from the name `reverse_input_channels` as it is not descriptive enough.

README.md
klemen1999: IMO conversion through YAML should be the preferred method since this is it's intention. Conversion through intermediate NNArchive is more like an added bonus since NNarchive itself holds only information about the model it is tied to and not about how to convert it to the next model - this is why we need to have assumptions built in when using intermediate NNArchive as input to the conversion (e.g. assumption that exported model should be BGR where as with config.yaml you can change this to RGB if you want)

README.md
```diff
@@ -101,9 +106,55 @@ To enable GPU acceleration for `hailo` conversion, install the [Nvidia Container
...
+- **Multi-Value `encoding.from` and `encoding.to`**:
+  Alternatively, you can explicitly set `encoding.from` and `encoding.to` to different values. For example:
+  ```yaml
+  encoding:
+    from: RGB
+    to: BGR
+  ```
+  This configuration specifies that the input data is in RGB format and will be converted to BGR format during processing.
+
+> \[!NOTE\] Note on Default Encoding
```
klemen1999: This isn't rendered correctly, I believe it should be `> [!NOTE]` (without `\`)? Same comment for other NOTEs

README.md
```diff
@@ -101,9 +106,55 @@ To enable GPU acceleration for `hailo` conversion, install the [Nvidia Container
...
+> \[!NOTE\] Note on Global Settings
+> Certain options can be set **globally**, applying to all inputs of the model, or **per input**. If specified per input, these settings will override the global configuration for that input alone. The options that support this flexibility include `scale_values`, `mean_values`, `encoding`, `data_type`, `shape`, and `layout`.
+
+#### NN Archive Configuration File
+
+In the NN Archive configuration, there are two flags related to color encoding control:
+
+- **`reverse_channels` (Deprecated)**:
+  Determines the input color format of the model: when set to *True*, the input is considered to be *"RGB"*, and when set to *False*, it is treated as *"BGR"*. This flag is deprecated and will be replaced by the `dai_type` flag in future versions.
+- **`dai_type`**:
```
klemen1999, line 148: Let's put this first (before `reverse_channels`) as it is the preferred method

README.md
```diff
@@ -101,9 +106,55 @@ To enable GPU acceleration for `hailo` conversion, install the [Nvidia Container
...
+  Provides a more comprehensive control over the input type compatible with the DAI backend. It is read by DepthAI to automatically configure the processing pipeline, including any necessary modifications to the input image format.
+
+> \[!NOTE\] Note on Default Settings
+> If neither `dai_type` nor `reverse_channels` the input to the model is considered to be *"BGR"*.
```
klemen1999: This is related to usecase where NNarchive is as intermediate right? In that case if neither is specified then the input to the model to be converted is considered to be "RGB".
This is already in the tests [here](https://github.com/luxonis/modelconverter/blob/230b1ffa84f3fe5d24973a07f871176a966c54f6/tests/test_utils/test_config.py#L683) so seems to be properly implemented already, just wrong here.

modelconverter/packages/rvc2/exporter.py
```diff
@@ -109,25 +109,50 @@ def _export_openvino_ir(self) -> Path:
...
                     inp.mean_values = inp.mean_values[::-1]
-                if inp.scale_values is not None and inp.reverse_input_channels:
+                if (
+                    inp.scale_values is not None
+                    and inp.encoding.from_ != inp.encoding.to
+                ):
                     inp.scale_values = inp.scale_values[::-1]
-                inp.reverse_input_channels = False
+                inp.encoding.from_ = Encoding.BGR
+                inp.encoding.to = Encoding.BGR
```
klemen1999, line 117: What is the reason for changing encodings here? And why do we set the both to BGR?

modelconverter/packages/hailo/exporter.py
```diff
@@ -237,7 +237,7 @@ def _get_alls(self, runner: ClientRunner) -> str:
                 f"{mean_values},{scale_values},{hn_name})"
             )
 
-            if inp.reverse_input_channels:
+            if inp.encoding.from_ != inp.encoding.to:
```
klemen1999: This check repeats alot and I agree with previous comments that we can make this into an internal property of InputConfig that just returns True/False depending if `from` and `to` are same

modelconverter/utils/onnx_tools.py
```diff
@@ -89,8 +89,12 @@ def onnx_attach_normalization_to_inputs(
             new_nodes.append(concat_node)
             last_output = f"normalized_{input_name}"
 
-        # 2. Subtract (mean) if mean_values is not None
-        if not reverse_only and cfg.mean_values is not None:
+        # 2. Subtract (mean) if mean_values is not None and not all 0
+        if (
+            not reverse_only
+            and cfg.mean_values is not None
```
klemen1999, line 95: By this point we need to make sure that if we are adding reverse channels (from and to are not the same) then mean_values and scale_values should also be already reversed. To test this E2E we could input dummy ONNX and check when blocks are added, when they aren't and what are their values.

klemen1999 (commented):
Generally looks good. I left some comments but mainly what we are missing is rigorous testing. By that I mean:
- test of the output NNArchive for different input configs, intermediate NNArchives and/or CLI params
- tests of intermediate onnx (for RVC4) to check which blocks are added and if values inside them are correct (mean and scale properly reversed if need be)
- test that flags for RVC2/3 and HAILO conversion are properly set up depending on the input config
- any other tests that will help catch any current or future bugs

.pre-commit-config.yaml
ptoupas: This change will be reverted before merging since its specific to my setup.

README.md
ptoupas: Correct that was a typo from my side.

README.md
ptoupas: Yeah that's right, apparently this is changed automatically by the `mdformat` of the `pre-commit`. Had to update README from github online editor.

modelconverter/packages/rvc2/exporter.py
ptoupas: This was added to be aligned with the previous implementation where `inp.reverse_input_channels = False` inside this if statement.
This is a special case where the model has >1 inputs and the inputs have different encodings (so one needs reverse of the channels and the other one does not). Since the `--reverse_input_channels` in the OpenVino is a universal one (e.g. applies to all inputs) what we are doing here is that we modify the onnx to add the necessary extra nodes to the inputs where channels need to be reversed and we make sure the `--reverse_input_channels` will not be set. To make that happen and due to this check [here](https://github.com/luxonis/modelconverter/blob/cbfdd2326aee31c7ad05ebea7cd95a38010e9ab3/modelconverter/packages/rvc2/exporter.py#L150-L155) we need to make sure that for all inputs the `encoding.from` and `encoding.to` will be the same and since the DAI default is BGR we explicitly set this to BGR.
Does this clear things out? Does this make sense?

modelconverter/packages/hailo/exporter.py
ptoupas: fixed on latest commit

tests/test_utils/test_config.py
```diff
@@ -552,6 +623,686 @@ def test_incorrect_type(key: str, value: str):
...
+def test_modified_onnx(keys: List[str], values: List[str]):
+    overrides = {keys[i]: values[i] for i in range(len(keys))}
+    overrides["input_model"] = str(DATA_DIR / "dummy_model.onnx")
+    config = Config.get_config(
+        None,
+        overrides,
+    )
+    inputs = next(iter(config.stages.values())).inputs
+    input_configs = {inp.name: inp for inp in inputs}
+    print(input_configs)
```
klemen1999: extra print

modelconverter/utils/config.py
kozlov721: > If you mean keeping reverse_input_channels as a flag in the config.yaml I don't agree with this.

No, that's not what I mean. It would be purely internal and not exposed to the users.

ptoupas: I think this is now the case with the [encoding_mismatch](https://github.com/luxonis/modelconverter/blob/50f4968a09014c8816b0d7a32e64fc6796fbc511/modelconverter/utils/config.py#L140-L142) property.

klemen1999 (approved):
LGTM, only thing left to update is the requirements.txt to luxonis-ml>=0.5.0
CC: @kozlov721 to re-review

kozlov721 (approved):
LGTM

PR #52, opened 2024-11-25
Hub cleanup, token storage, readme

kozlov721:
Various small changes:
- Some CLI cleanup
- Updated README
- Added `modelconverter hub login` command to securely store the API key on the system 
  - So users don't need to figure out env variables for online conversion


modelconverter/cli/types.py
```diff
@@ -1,38 +1,13 @@
...
 from typing import List, Optional
 
+import click
 import typer
 from typing_extensions import Annotated
 
 from modelconverter.utils.types import Target
 
 
-class Task(str, Enum):
```
klemen1999, line 11: What is the reason for dropping these enums? Because they are shown by click choices? From code quiality standpoint I feel like enums do help keep everything more organized.

modelconverter/hub/__main__.py
```diff
@@ -1,12 +1,17 @@
 import logging
+import webbrowser
 from pathlib import Path
-from typing import Any, Dict, List, Literal, Optional
+from time import sleep
+from typing import Any, Dict, List, Optional
 from urllib.parse import unquote, urlparse
 
+import click
+import keyring
```
klemen1999, line 9: Do these 2 need to be added to requirements?

klemen1999 (approved):
LGTM, just left some comments

modelconverter/hub/__main__.py
kozlov721: Only `keyring` byt yes, thanks

modelconverter/cli/types.py
kozlov721: tldr; they're only used as a way to limit some CLI arguments because `typer` doesn't support the `Literal` type. 

---

`typer` that we're using to build the CLI doesn't support the `Literal` type in function annotations. The only way how to get a nice support for an argument with a set of restricted values is to create an enum, so the enums are only used for this. This makes using the functions outside of the CLI a bit cumbersome (which started to be an issue with the online conversion). 
 
There's an open issue and even a PR to add the `Literal` support for several years but it seems somehow ignored. However, one of the comments there showed how to achieve a bit similar behavior to `Literal` by leveraging the underlying `click` library that `typer` is built on, which I found a bit nicer to work with. 

Alternative way would be to have two sets of functions. One that would be more simple to use programmatically (with `Literal`s) and an almost identical set of signatures with enums for the CLI (implementation-wise thye'd be just wrappers around the first set). From the user perspective this would probably be the cleanest, but it would be a bit ugly codebase-wise imo.

klemen1999: Ok I see I see.. yeah in that case let's leave it as is

PR #55, opened 2024-12-10
Add ONNXModifier for optimising the ONNX model before converting for RVC4 execution

ptoupas:
The ONNXModifier performs a series of onnx modifications that were found to be beneficial in terms of performance on RVC4, without altering the final outputs of the model. Below is the list of optimisations/modifications introduced:
- Substituting division layers for multiplication layers (when possible).
- Substituting subtraction layers for addition layers (when possible).
- Fusing add/mul layers that come right after a conv layer, into a BatchNormalization layer, which is then fused into the conv (when possible).
- Fusing add/mul layers that precede a conv layer into the conv layer by incorporating the addition and multiplication into the conv's weights and biases when possible.
- Removing the Split and Concat layers at the start of the model by reordering the covolutional weights in first conv layer when such transformation is possible.


modelconverter/utils/onnx_tools.py
```diff
@@ -176,3 +181,975 @@ def onnx_attach_normalization_to_inputs(
 
     onnx.save(model, str(save_path))
     return save_path
+
+
+@dataclass
```
kozlov721: IMHO in this situation using `dataclass` is making it more complicated than using plain python class.

modelconverter/packages/rvc4/exporter.py
```diff
@@ -57,6 +59,19 @@ def __init__(self, config: SingleStageConfig, output_dir: Path):
                 self._attach_suffix(self.input_model, "modified.onnx"),
                 self.inputs,
             )
+
+            onnx_modifier = ONNXModifier(
```
kozlov721: Would these modifications be beneficial in general for all the target platforms? If so, we could move the modification up from here to base `Exporter`.

kozlov721: We should also add an option to the config to disable the modifications

modelconverter/utils/onnx_tools.py
```diff
@@ -176,3 +181,975 @@ def onnx_attach_normalization_to_inputs(
...
+    def __post_init__(self):
+        self.load_onnx()
+
+    def load_onnx(self) -> None:
+        """Load the ONNX model and store it as onnx.ModelProto and
+        onnx_graphsurgeon.GraphSurgeon graph."""
+
+        logger.info(f"Loading model: {self.model_path.stem}")
+
+        # Load the ONNX model
```
kozlov721: Most/all of the inline comments can be deleted. It's pretty clear what's happening from the code.

modelconverter/utils/onnx_tools.py
```diff
@@ -1,10 +1,15 @@
 import logging
+from dataclasses import dataclass
 from pathlib import Path
-from typing import Dict
+from typing import Dict, List, Optional, Tuple
 
+import numpy as np
 import onnx
+import onnx_graphsurgeon as gs
```
kozlov721, line 7: Could we use `graphsurgeon` to also simplify the code in `onnx_attach_normalization_to_inputs`? Or add it to the `ONNXModifier`

requirements-dev.txt
```diff
@@ -1,3 +1,5 @@
 pre-commit==3.2.1
 pytest-cov>=4.1.0
 docker-squash>=1.1.0
+onnx_graphsurgeon
```
kozlov721: The dev requirements don't need to be updated. They're installed on top of the regular requirements.

kozlov721 (commented):
Thanks! LGTM in general, just left a few comments.

modelconverter/packages/rvc4/exporter.py
ptoupas: These modifications were found to be beneficial for the RVC4 case after some tests and benchmarking @aljazkonec1 did. I think its unknown if these will improve (or worsen) the performance on the rest of the devices. Maybe @aljazkonec1 has a better view on this.

modelconverter/utils/onnx_tools.py
ptoupas: Even though I would like the `onnx_attach_normalization_to_inputs` to be another function of `ONNXModifier`, I think its better to leave this one outside of the class. The reason is that its the fallback solution in case the optimisations fail or do not produce the correct results compared to the "base" model which is produced by this exact function.

ptoupas: But the `onnx_attach_normalization_to_inputs` alone can be changed to use `graphsurgeon` instead, yeah.

modelconverter/packages/rvc4/exporter.py
aljazkonec1: I did compare output values between original onnx and the optimized onnx. In my limited testing, there were no differences between original onnx and the optimized onnx outputs. From the graph it was mostly operations like split, concat, reshape that were simplified and fused into other layers like conv

modelconverter/utils/onnx_tools.py
```diff
@@ -176,3 +181,975 @@ def onnx_attach_normalization_to_inputs(
...
+        1. Conv -> Add -> Mul
+        2. Conv -> Sub -> Mul
+        3. Conv -> Mul -> Add
+        4. Conv -> Mul -> Sub
+        5. Conv -> Mul
+        6. Conv -> Add
+        7. Conv -> Sub
+        """
+
+        FUSION_PATTERNS = [
```
aljazkonec1, line 475: I would only keep the following patterns: 
(Conv, Add, Mul)
(Conv, Mul, Add)
(Conv, Mul)
(Conv, Add)

There should be no sub or div layers when running this function due to the replace function that is ran before. This would make it easier for debugging later and there are less options to take care of in this file. 

Potentially I would even move (Conv, add) and (Conv, Mul) to `fuse_single_add_mul_to_conv` as there we also fuse single layers instead of making bn layers out of these.

modelconverter/utils/onnx_tools.py
```diff
@@ -176,3 +181,975 @@ def onnx_attach_normalization_to_inputs(
...
+        self.substitute_node_by_type(source_node="Sub", target_node="Add")
+        logger.debug(
+            "Fusing Add and Mul nodes to BatchNormalization nodes and then into Conv nodes..."
+        )
+        self.fuse_add_mul_to_bn()
+        logger.debug("Fusing Add and Mul nodes to Conv nodes...")
+        self.fuse_comb_add_mul_to_conv()
+        self.fuse_single_add_mul_to_conv()
+        logger.debug("Fusing Split and Concat nodes to Conv nodes...")
+        self.fuse_split_concat_to_conv()
```
aljazkonec1: I think It would be beneficial to run `compare_outputs` function after every modification is done. And have the logger / converter give a big red warning to which stage the model was modified to. This would also help with faster debugging for internal use.

aljazkonec1 (commented):
Looks good, some comments mostly regarding where and what operations to check.

modelconverter/packages/rvc4/exporter.py
aljazkonec1: Yes, the optimizations were done on the basis of what was slowing inference speeds on RVC4 devices. My understanding is that converting to RVC2 is pretty optimized by default from Intel itself and we do not have much insights into per layer analysis of models. In any case, this PR only affects add, mul, div, sub layers and fusing them into already existing Conv layers. I do not see a reason that different weights in Conv layers would affect inference speed on RVC2

klemen1999: Actually RVC2/Intelalso has a [benchmarking tool](https://docs.openvino.ai/2024/learn-openvino/openvino-samples/benchmark-tool.html) which we could use to check how/if these optimizations affect the RVC2 exported model. But IMO we can leave this for a separate PR because we want to test it out first to verify.
+1 on optionally disabling modifications through config (by default I would have it on)

modelconverter/utils/onnx_tools.py
klemen1999: +1, easier potential debugging outways the small addition in time needed for conversion IMO

klemen1999 (approved):
LGTM

modelconverter/utils/onnx_tools.py
```diff
@@ -388,69 +387,55 @@ def substitute_node_by_type(
...
         constant_map = self.get_constant_map(self.onnx_gs)
 
         def create_new_node(
             node: gs.Node, target_node: str, const_idx: int
-        ) -> Optional[List[gs.Node]]:
+        ) -> Optional[gs.Node]:
             if const_idx == 0:
                 return None
-            new_nodes = []
+            # new_nodes = []
```
klemen1999: we can remove

klemen1999 (approved):
LGTM

PR #56, opened 2024-12-10
Updated ONNX version

kozlov721:
Updating `onnx` package to version `1.17` to fix a discovered [vulnerability](https://github.com/advisories/GHSA-6rq9-53c3-f7vj)

### DIFF
--- a/modelconverter/utils/environ.py
+++ b/modelconverter/utils/environ.py
@@ -3,12 +3,14 @@ from typing import Optional
 
 import keyring
 from luxonis_ml.utils import Environ as BaseEnviron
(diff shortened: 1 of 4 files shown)


klemen1999 (approved):
LGTM but let's check if this also affects anything in the #55 CC: @ptoupas

ptoupas:
@kozlov721 I think there might be an issue with the #55 if we are to move to `1.17` version.
Actually it does not have to do with #55 directly, its just that the `test_config.py` wont pass with `onnx=1.17` version (at least this happens locally to me).

PR #64, opened 2025-01-09
Update RVC2 and RVC4 benchmark scripts to work with the dai Benchmark Nodes

ptoupas:
- Updated the RVC2 benchmark to work with dai V3 using the newly added Benchmark Nodes.
   - The new script accepts `.blob`, `.tar.xz`, and model slugs from HubAI.
- Added a second benchmark option to RVC4 to use dai V3 as the way to bechmark the models.
   - The new script accepts `.tar.xz`, and model slugs from HubAI. It does not accept `.dlc` files.

Using the dai Benchmark Nodes the latency measurements are not valid in most cases, so this is omitted from the Benchmark reports in the case where the dai benchmark is being used.


modelconverter/utils/hubai_utils.py
```diff
@@ -0,0 +1,36 @@
+import requests
+
+
+def is_hubai_available(model_slug: str) -> bool:
```
klemen1999: Instead of this function can we use [this one](https://github.com/luxonis/modelconverter/blob/89852a47d82cc2bb43eb2051060e44857165e824/modelconverter/cli/utils.py#L302)? Coupled with `try except` it should produce the same thing with cleaner code I believe?

klemen1999 (approved):
LGTM thanks!

modelconverter/utils/hubai_utils.py
klemen1999: Actually the function has exit(1). Might have to use only the [Request.get()](https://github.com/luxonis/modelconverter/blob/89852a47d82cc2bb43eb2051060e44857165e824/modelconverter/cli/utils.py#L309) part and not throw exit() on except

modelconverter/utils/hubai_utils.py
```diff
@@ -1,36 +1,27 @@
...
+        "modelVersions/", params={"model_id": model_id, "is_public": True}
+    )
+
+    for version in model_variants:
         if (
-            slug in model_slug
-            or slug.removeprefix(f"{model['team_slug']}/") in model_slug
+            f"{model_name}:{version['variant_slug']}" == model_slug
+            or f"{model_name}:{version['variant_slug']}".removeprefix(
+                "luxonis/"
```
klemen1999: The prefix is team specific so it shouldn't be hardoced to "luxonis". We can first parse the whole slug which is in format  `<team_name>/<model name>:<variant>` and then dynamically use specific parts

kozlov721 (approved):

klemen1999 (approved):
LGTM!

PR #70, opened 2025-02-20
Fix: hailo tests

kozlov721:
## Purpose

Fixes failing `hailo` workflows in GH Actions.
## Specification

Renamed `docker/hailo/Dockerfile.internal` to `docker/hailo/Dockerfile` before running the tests.


klemen1999 (approved):
IMO we can merge to not block release and address hailo conversion in next one

PR #71, opened 2025-02-20
Fix: Keyring and Hub Request 

kozlov721:
## Purpose

Fix for hanging `keyring` operations on certain systems.

## Specification

- Added timeout for retrieving API key from the keyring
- Improved the requests state so the key can be changed dynamically
- Improved error messages

### DIFF
--- a/modelconverter/utils/environ.py
+++ b/modelconverter/utils/environ.py
@@ -4,8 +4,8 @@ from typing import Optional, Union
 
 import keyring
 from luxonis_ml.utils import Environ as BaseEnviron
-from pydantic import AliasChoices, Field, model_validator
-from typing_extensions import Annotated, Self
+from pydantic import model_validator
+from typing_extensions import Self
 
 
 def get_password_with_timeout(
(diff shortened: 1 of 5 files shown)


modelconverter/utils/environ.py
```diff
@@ -1,25 +1,47 @@
...
+        except TimeoutError:
+            return None
+
+
 class Environ(BaseEnviron):
     HUBAI_API_KEY: Annotated[
-        Optional[str], Field(validation_alias=AliasChoices("HUB_AI_API_KEY"))
+        Optional[str],
+        Field(
+            validation_alias=AliasChoices("HUBAI_API_KEY", "HUB_AI_API_KEY")
```
klemen1999: Is there any specific reason we want to accept both (same for Hubai_url)? I would opt from `HUBAI` being one word and have only that if there is no need for two. I think we just need to document which one is used.

klemen1999 (approved):
LGTM, just one comment

PR #75, opened 2025-02-24
Downloaded models in `test-modifier`

kozlov721:
## Purpose

Reduces the number of pulled models in `test-modifier`

### DIFF
--- a/tests/test_utils/test_modifier.py
+++ b/tests/test_utils/test_modifier.py
@@ -34,7 +34,7 @@ def download_onnx_models():
         os.makedirs(DATA_DIR)
 
     hub_ai_models = Request.get(
-        "models/", params={"is_public": True, "limit": 1000}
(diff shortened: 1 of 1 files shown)


tests/test_utils/test_modifier.py
```diff
@@ -34,7 +34,7 @@ def download_onnx_models():
         os.makedirs(DATA_DIR)
 
     hub_ai_models = Request.get(
-        "models/", params={"is_public": True, "limit": 1000}
+        "models/", params={"is_public": True, "limit": 10}
```
klemen1999, line 37: Can we query which are public and then randomly pick 10 instead of having same 10 each time?

kozlov721: This is more of just a hotfix; I'll eventually create a dummy model with all the layers we modify for this test.

klemen1999 (approved):

PR #77, opened 2025-03-12
Add hub CLI YOLO models converter

sokovninn:
## Purpose
Added support for converting YOLO models using the hub CLI.

## Specification
Introduced CLI options for YOLO input shape, version, and class names. Updated model validation to handle YOLO models and enforce required parameters.

No breaking changes.

## Deployment Plan
Standard rollout with existing deployment pipeline.

## Testing & Validation
Tested via CLI with different YOLO versions and input configurations.

### DIFF
--- a/modelconverter/hub/__main__.py
+++ b/modelconverter/hub/__main__.py
@@ -8,6 +8,7 @@ import click
 import keyring
 import requests
 import typer
+from loguru import logger
 from luxonis_ml.nn_archive import is_nn_archive
 from rich import print
 from typing_extensions import Annotated
@@ -58,6 +59,9 @@ from modelconverter.cli import (
     TasksOption,
     VariantSlugOption,
     VersionOption,
+    YoloClassNamesOption,
+    YoloInputShapeOption,
+    YoloVersionOption,
     get_configs,
     get_resource_id,
     get_target_specific_options,
@@ -513,6 +517,8 @@ def _export(
     target: str,
     target_precision: str,
     quantization_data: str,
+    yolo_version: Optional[str] = None,
+    yolo_class_names: Optional[List[str]] = None,
     **kwargs,
 ) -> Dict[str, Any]:
     """Exports a model instance."""
@@ -522,6 +528,14 @@ def _export(
         "quantization_data": quantization_data,
         **kwargs,
     }
+    if yolo_version:
+        json["version"] = yolo_version
+    if yolo_class_names:
+        json["class_names"] = yolo_class_names
+    if yolo_version and not yolo_class_names:
+        logger.warning(
+            "YOLO class names are required for YOLO models. Please provide them using `--yolo-class-names`. Otherwise, default class names will be used."
+        )
     if target == "RVC4":
         json["target_precision"] = target_precision
     res = Request.post(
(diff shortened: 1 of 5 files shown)


modelconverter/utils/config.py
```diff
@@ -302,7 +302,34 @@ def _validate_model(cls, data: Dict[str, Any]) -> Dict[str, Any]:
...
+                input_shape = (
+                    list(map(int, input_shape.split(" ")))
+                    if " " in input_shape
+                    else [int(input_shape)] * 2
+                )
+            else:
+                logger.warning(
+                    "yolo_input_shape is not provided. Using default shape [640, 640]."
+                )
+            input_shapes = {"images": input_shape}
```
HonzaCuhel: Height must be passed first, not width. Please replace with `input_shapes = {"images": input_shape[::-1]}`.

modelconverter/hub/__main__.py
```diff
@@ -522,6 +528,14 @@ def _export(
         "quantization_data": quantization_data,
         **kwargs,
     }
+    if yolo_version:
+        json["version"] = yolo_version
+    if yolo_class_names:
+        json["class_names"] = yolo_class_names
+    if yolo_version and not yolo_class_names:
+        logger.warning(
+            "YOLO class names are required for YOLO models. Please provide them using `--yolo-class-names`. Otherwise, default class names will be used."
```
HonzaCuhel: I'd formulate it a bit, I wouldn't use the word required as it is not really required. Tools-cli tries to retrieve the class names from the model weights, and some models actually have this information stored, so I'd formulate it like this:
```
"It's recommended to provide YOLO class names via --yolo-class-names. If omitted, class names will be extracted from model weights if present, otherwise default names will be used."
```

HonzaCuhel (commented):
Generally, LGTM, but I left some notes.

If the `yolo_version` isn't provided, the conversion fails, as this is a required argument by HubAI BE. The way I see it, there are two possible solutions:

1. Quick, but not optimal IMO: if pytorch weights are passed then the `yolo_version` argument is forced.
2. More complex, but IMO better solution: use the [tools-cli version detector](https://github.com/luxonis/tools/blob/main/tools/utils/version_detection.py#L22) to retrieve the YOLO version based on the model weights and use it when the `yolo_version` parameter wasn't specified and if specified, check it's the same version. The motivation is that it'll be more convenient to the users this way since it's one thing less they'd need to worry about.

Thoughts on this @sokovninn, @klemen1999, @kozlov721?

klemen1999:
> If the yolo_version isn't provided, the conversion fails, as this is a required argument by HubAI BE. The way I see it, there are two possible solutions:

IMO for this first iteration we make this a required parameter. And we can then think of integrating the `detect_version` as a separate feature. We would likely want to restructure this a bit before on the `tools` side becuse the whole repo dependency brings quite a lot of other, not needed stuff. But we could split it into modules so only e.g. detect module would need to be installed in modelconverter.

modelconverter/utils/config.py
sokovninn: I took the width-first format from tools-cli. But HubAI BE requires height first, right?

HonzaCuhel: It’s true that tools-cli uses width-first, but HubAI BE ruses height-first.

HonzaCuhel (approved):

PR #78, opened 2025-03-14
Fix allowed number of threads for RVC4 benchmark

ptoupas:
## Purpose
Update RVC4 benchmark to be able to work with threads > 1.

## Specification
Removed the respective code that was forcing the RVC4 benchmark to only run on a single thread due to some issues with DAI library.

No breaking changes.

## Deployment Plan
Standard rollout with existing deployment pipeline.

## Testing & Validation
Tested with various models on different RVC4 devices

### DIFF
--- a/requirements-bench.txt
+++ b/requirements-bench.txt
@@ -1,2 +1,2 @@
-depthai>=3.0.0a12
+depthai>=3.0.0a13
 pandas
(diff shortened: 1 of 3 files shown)


requirements-bench.txt
```diff
@@ -1,2 +1,2 @@
-depthai>=3.0.0a12
+depthai>=3.0.0a13
```
klemen1999: Let's use a14
```suggestion
depthai>=3.0.0a14
```

klemen1999 (approved):
LGTM

PR #82, opened 2025-03-31
Add Analyzer and Visualizer packages to modelconverter

aljazkonec1:
## Purpose

This PR adds analysis tools to help validate a DLC model is successfully converted & quantized. The results of multiple models can then be visualized in a pyplot plot for fast and easy comparison. The tools rely on the user already having converted a model / models to DLC and wants to see the affect of quantization and layer speeds on the device.

The most general way to use the analyzer is simply:
`modelconverter analyze dlc_model onnx_model image_dir`
This will result in two CSV files, one containing statistics on layer outputs compared to the base ONNX model and the second containg cycle usage on an RVC4 device.

Once one or more  models are converted, the results can be visualized with the use of:
`modelconverter visualize` which will automatically search for all relevant CSV files in `shared_with_container/outputs/analysis` and create comparison plots for them. This greatly simplifies and speeds up validation of a converted model. The tools work with as little as one image in a directory, but for more accurate results, I recomend around 50 images.

## Specification

Added Analayzer and Visualizer classes to RVC4 package, added new CLI commands ( `analyze` , `analyze-outputs`, `analyze-cycles`, `visualize`).

Does not affect other packages as the assumption was that the model is already converted. There are some assumptions about the model:
-  all ONNX models are in NCHW format and all DLC model outputs are in NHWC format. 
- Inputs to the models are BGR images in 0-255 format (even if it is then saved as float32)

## Deployment Plan

Further development will be made in three areas:
- A logger will be added to monitor status
- A better performance metric. I still need to think about what is the best option
- Updated visuals that give a better, deeper understanding of the model

## Testing & Validation

Tested on multiple yolo models, and tested previously in another directory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,23 +23,27 @@ Convert your **ONNX** models to a format compatible with any generation of Luxon
 
 ## Table of Contents
 
-- [Installation](#installation)
-- [Configuration](#configuration)
-  - [YAML Configuration File](#yaml-configuration-file)
-  - [NN Archive Configuration File](#nn-archive-configuration-file)
-- [Online Usage](#online-usage)
-- [Local Usage](#local-usage)
-  - [Prerequisites](#prerequisites)
-    - [GPU Support](#gpu-support)
-  - [Sharing Files](#sharing-files)
-  - [Running ModelConverter](#running-modelconverter)
-    - [Examples](#examples)
-- [Multi-Stage Conversion](#multi-stage-conversion)
-- [Interactive Mode](#interactive-mode)
-- [Calibration Data](#calibration-data)
-- [Inference](#inference)
-  - [Inference Example](#inference-example)
-- [Benchmarking](#benchmarking)
+- [ModelConverter - Compilation Library](#modelconverter---compilation-library)
+  - [Status](#status)
+  - [Table of Contents](#table-of-contents)
+  - [Installation](#installation)
+  - [Configuration](#configuration)
+    - [YAML Configuration File](#yaml-configuration-file)
+    - [NN Archive Configuration File](#nn-archive-configuration-file)
+  - [Online Usage](#online-usage)
+  - [Local Usage](#local-usage)
+    - [Prerequisites](#prerequisites)
+      - [GPU Support](#gpu-support)
+    - [Sharing Files](#sharing-files)
+    - [Running ModelConverter](#running-modelconverter)
+      - [Examples](#examples)
+  - [Multi-Stage Conversion](#multi-stage-conversion)
+  - [Interactive Mode](#interactive-mode)
+  - [Calibration Data](#calibration-data)
+  - [Inference](#inference)
+    - [Inference Example](#inference-example)
+  - [\[RVC4\] DLC model analysis](#rvc4-dlc-model-analysis)
+  - [Benchmarking](#benchmarking)
 
 ## Installation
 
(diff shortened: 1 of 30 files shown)


aljazkonec1:
The visualizations are interactive (you filter shown models, hover over for more info, change metric, zoom in) and should look somehting like:
image
image

modelconverter/__main__.py
```diff
@@ -224,6 +230,112 @@ def benchmark(
...
+
+    Analyzer = get_analyzer(Target.RVC4)
+    analyzer = Analyzer(dlc_model_path, image_dirs_dict)
+
+    analyzer.analyze_layer_outputs(resolve_path(onnx_model_path, Path.cwd()))
+    analyzer.analyze_layer_cycles()
+
+
+@app.command()
+def analyze_outputs(
```
ptoupas: I suggest moving this as an "internal choice"/flag inside the `analyze` command instead of having this as a command on its own.
This way the CLI interface will remain more "clear" and also will follow the way the rest of the commands work.
And based on the flag do both `analyze_layer_outputs` and `analyze_layer_cycles` or just one of them.

modelconverter/__main__.py
```diff
@@ -224,6 +230,112 @@ def benchmark(
...
+        }
+
+    Analyzer = get_analyzer(Target.RVC4)
+    analyzer = Analyzer(dlc_model_path, image_dirs_dict)
+
+    analyzer.analyze_layer_outputs(resolve_path(onnx_model_path, Path.cwd()))
+
+
+@app.command()
+def analyze_cycles(
```
ptoupas: Same comment as above for `analyze_outputs`.

README.md
```diff
@@ -23,23 +23,27 @@ Convert your **ONNX** models to a format compatible with any generation of Luxon
...
-    - [Examples](#examples)
-- [Multi-Stage Conversion](#multi-stage-conversion)
-- [Interactive Mode](#interactive-mode)
-- [Calibration Data](#calibration-data)
-- [Inference](#inference)
-  - [Inference Example](#inference-example)
-- [Benchmarking](#benchmarking)
+- [ModelConverter - Compilation Library](#modelconverter---compilation-library)
+  - [Status](#status)
+  - [Table of Contents](#table-of-contents)
```
ptoupas: Can we remove these three from the ToC to keep the structure as it was previously?

requirements-analysis.txt
```diff
@@ -0,0 +1,2 @@
+pandas
```
ptoupas: Since we are moving away from pandas in modelconverter and replacing that with polars, could we change this requirement and also update any code that uses pandas to use polars instead?
You can have a look at the respective changes on [this](https://github.com/luxonis/modelconverter/commit/29dd45889a4e257d7813110187a40bf2dc20706c) commit.

README.md
```diff
@@ -428,6 +432,58 @@ output_path/
...
+
+ModelConverter offers additional analysis tools for the RVC4 platform. The tools provide an in-depth look at the following:
+
+1. The outputs of all layers in comparison to the ground truth ONNX model,
+1. The cycle usage of each layer on an RVC4 device.
+1. Visualizations for fast and easy comparison of multiple models.
+
+This gives the user better insight into the successful quantization of a model, helps discover potential speed bottleneck layers, and allows for the comparison of different quantization parameters.
+
+To install the package with the analysis dependencies, use:
```
ptoupas, line 481: Could we also add some comment on the SNPE tools dependencies?
I assume the SNPE tools must be installed locally (on the host machine) for this to work?
If this is not needed, ignore this comment.

modelconverter/__main__.py
```diff
@@ -224,6 +230,112 @@ def benchmark(
...
+        image_dirs_dict = {"default": image_dirs[0]}
+    else:
+        if len(image_dirs) % 2 != 0:
+            raise typer.BadParameter(
+                "Please supply the same amount of model input names and test image directories."
+            )
+        image_dirs_dict = {
+            image_dirs[i]: image_dirs[i + 1]
+            for i in range(0, len(image_dirs), 2)
+        }
```
ptoupas: Can we also add a check here that all the `image_dirs` have the exact same amount of images and that the names on all the (sorted) images match?

modelconverter/packages/rvc4/analyze.py
```diff
@@ -0,0 +1,341 @@
...
+        input_matcher = {}
+        for i in range(len(next(iter(image_names.values())))):
+            input_matcher[i] = {}
+            for k in image_names.keys():
+                input_matcher[i][k] = str(image_names[k][i])
+
+        # input_matcher = { 0 : {input_name1 : full_path_to_img1, input_name2 : full_path_to_img2},
+        #                   1 : {input_name1 : full_path_to_img2, input_name2 : full_path_to_img2},
+        #                   ....
+        #                   n : {input_name1 : full_path_to_imgn, input_name2 : full_path_to_imgn}}
```
ptoupas: Should we remove these comments?

modelconverter/packages/rvc4/analyze.py
```diff
@@ -0,0 +1,341 @@
...
+    def __init__(self, dlc_model_path: str, image_dirs: Dict[str, str]):
+        super().__init__(dlc_model_path, image_dirs)
+        self.adb = AdbHandler()
+
+    def analyze_layer_outputs(self, onnx_model_path: Path) -> None:
+        input_matcher = self._prepare_input_matcher()
+        dlc_matcher = self._prepare_raw_inputs(input_matcher, np.float32)
+
+        output_dir = self._run_dlc(
+            f"snpe-net-run --container {self.model_name}.dlc --input_list input_list.txt --debug --use_dsp --userbuffer_floatN_output 32 --perf_profile balanced --userbuffer_float"
```
ptoupas, line 29: I am just wondering what if the dlc requires INT8 inputs? If I correctly understand this will execute the dlc with float inputs?
I guess I might be missing something here, could you please provide some more details on that?

ptoupas (changes requested):
Overall LGTM. I've left some comments that needs to be addressed.
I've got one more general question though, @aljazkonec1. If I am not mistaken this analysis only works on models that have images as inputs right?
Do we want to extend that so we can have this analysis on other models as well with non-image inputs?
Probably not very high priority, just bringing that question up. Thoughts @klemen1999?

README.md
aljazkonec1: All tools should already be installed when using `modelconverter` CLI from PyPI (tools are in the zip in RVC4 section of[Prerequisites](https://github.com/luxonis/modelconverter/tree/main?tab=readme-ov-file#prerequisites))

ptoupas: So everything is done inside the docker container when it comes to `Analyzer` and `Visualizer`?

aljazkonec1: yes, it uses same logic as for example exporter

modelconverter/packages/rvc4/analyze.py
aljazkonec1: It still works because SNPE automatically converts it. I tested with yolo and ocr models that both have INT8 inputs. Its just that if we use INT8 images, then the output will be in INT8 (the outputs need to be unquantized)

ptoupas: Oh, alright then, I didn't know about this SNPE internal conversion. If that's the case, this should also be considered resolved.

ptoupas:
Generally LGTM.
I think that before merging we need to add the docker related code so that the [analyze](https://github.com/luxonis/modelconverter/blob/0c272eefe1353417756b3be9467b7e0840cb5aed/modelconverter/__main__.py#L235-L242) CLI command is executed inside docker and not on the host (same as the [convert](https://github.com/luxonis/modelconverter/blob/0c272eefe1353417756b3be9467b7e0840cb5aed/modelconverter/__main__.py#L302-L325) cli).
Otherwise there are extra requirements for this to work and these are the SNPE tools being installed on the host machine.
What do you think @klemen1999, @kozlov721?

ptoupas (approved):

PR #83, opened 2025-04-03
Remove pandas dependency and interchange that with polas.

ptoupas:
## Purpose
This PR removes the dependencies of the modelconverter from pandas and changes over to polar. The reason of this change is because polars is an overall faster library and it is already incorporated into luxonis-ml which is being used by this tool.

## Specification
Changed benchmark related scripts that have been using pandas such as [base_benchmark.py](https://github.com/luxonis/modelconverter/blob/main/modelconverter/packages/base_benchmark.py), [benchmark.py](https://github.com/luxonis/modelconverter/blob/main/modelconverter/packages/rvc4/benchmark.py), as well as some utils files.

Does affect the modelconverter benchmark and convert tools.

## Deployment Plan
None / not applicable

## Testing & Validation
Tested the change from pandas to polars on all cases where it was used by checking 1 to 1 the outputs of the pandas and polars implementations on different inputs (mostly csv files).


modelconverter/packages/rvc4/benchmark.py
```diff
@@ -126,17 +126,14 @@ def _get_input_sizes(self) -> Tuple[Dict[str, List[int]], Dict[str, str]]:
 
         # Extract and load the relevant CSV part into a pandas DataFrame.
```
klemen1999:
```suggestion
        # Extract and load the relevant CSV part into a polars DataFrame.
```

klemen1999 (approved):
LGTM, @JSabadin could you check as well just to be sure this move to polars doesn't introduce any issues with Windows? (it shouldn't but better to check :))

PR #112, opened 2025-06-16
RVC4: Simpler support for FP16 export

kozlov721:
## Purpose

Makes it easier to convert RVC4 models to FP16 precision.

## Specification

- Added new configuration field `rvc4.compress_to_fp16` which will automatically add the required flags to `snpe-onnx-to-dlc` and `snpe-dlc-graph-prepare`

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable


ptoupas (approved):
LGTM.
Just one minor comment. In case of the `rvc4.compress_to_fp16 True` flag is used I would remove the second warning about the calibration that has been disabled since its already stated in the warning above (see attached figure).

image

PR #113, opened 2025-06-17
ONNXModifier simplification and optimization fixes

ptoupas:
## Purpose
Fixes some issues during the optimization and simplification of the ONNX models during conversion.

## Specification
- Update the IR version if not compatible with the expected version 10.
- Use `onnxruntime` for the model's optimizations instead of `onnxoptimizer`. 
- Small fixes on edge cases with new nodes we create and naming of those

None / not applicable

## Deployment Plan
None / not applicable

## Testing & Validation
Tested on a few models such as `zero-dce`, and `l2cs` previously failing to convert due to issues in the optimization and simplification of the ONNX model graph.

### DIFF
--- a/modelconverter/utils/onnx_tools.py
+++ b/modelconverter/utils/onnx_tools.py
@@ -292,7 +292,7 @@ class ONNXModifier:
                     sess_options.optimized_model_filepath
                 )
 
-                Path.unlink(Path(sess_options.optimized_model_filepath))
+                Path(sess_options.optimized_model_filepath).unlink()
 
         optimized_onnx_model, _ = simplify(
             optimized_onnx_model, perform_optimization=False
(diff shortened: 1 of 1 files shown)


modelconverter/utils/onnx_tools.py
```diff
@@ -254,19 +261,38 @@ def load_onnx(self) -> None:
...
+                    tmp_file.name + "_optimized.onnx"
+                )
+
+                _ = ort.InferenceSession(tmp_file.name, sess_options)
+
+                optimized_onnx_model = onnx.load(
+                    sess_options.optimized_model_filepath
+                )
+
+                Path.unlink(Path(sess_options.optimized_model_filepath))
```
kozlov721: This can be `Path(sess_options.optimized_model_filepath).unlink()`

kozlov721 (approved):
LGTM, thanks

PR #116, opened 2025-06-18
Fix: NN Archive precision

kozlov721:
## Purpose

Fixes incorrectly set `precision` and `dtype` fields when converting to an NN Archive.

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/modelconverter/utils/nn_archive.py
+++ b/modelconverter/utils/nn_archive.py
@@ -208,12 +208,28 @@ def modelconverter_config_to_nn(
     target_cfg = cfg.get_target_config(target)
 
     # TODO: This might be more complicated for Hailo
-    if target is Target.RVC2 or target_cfg.disable_calibration:
-        precision = DataType.FLOAT32
-    elif getattr(target_cfg, "compress_to_fp16", False):
-        precision = DataType.FLOAT16
-    else:
-        precision = DataType.INT8
+
+    compress_to_fp16 = getattr(target_cfg, "compress_to_fp16", False)
+    disable_calibration = target_cfg.disable_calibration
+
+    match target, compress_to_fp16, disable_calibration:
+        case Target.RVC2, True, _:
+            precision = DataType.FLOAT16
+
+        case Target.RVC2, False, _:
+            precision = DataType.FLOAT32
+
+        case Target.RVC3 | Target.RVC4, True, True:
+            precision = DataType.FLOAT16
+
+        case Target.RVC3 | Target.RVC4, False, True:
+            precision = DataType.FLOAT32
+
+        case Target.RVC3 | Target.RVC4, _, False:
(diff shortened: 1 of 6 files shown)


modelconverter/utils/nn_archive.py
```diff
@@ -199,18 +199,29 @@ def modelconverter_config_to_nn(
...
     is_multistage = len(config.stages) > 1
     model_metadata = get_metadata(model_path)
 
     cfg = config.stages[main_stage_key]
+    target_cfg = cfg.get_target_config(target)
+
+    # TODO: This might be more complicated for Hailo
+    if target_cfg.disable_calibration:
+        precision = DataType.FLOAT32
+    elif getattr(target_cfg, "compress_to_fp16", False):
```
klemen1999: With RVC2 models this check gets hit by default right? Just double checking.
But what if we set compress_to_fp16 false with rvc? Right now it would output INT8 precision where is I think the OV in that case uses FP32? (not 100% sure on this, needs to be checked)

modelconverter/utils/nn_archive.py
```diff
@@ -238,7 +249,9 @@ def modelconverter_config_to_nn(
                 "name": inp.name,
                 "shape": new_shape,
                 "layout": layout,
-                "dtype": inp.data_type.value,
+                "dtype": model_metadata.input_dtypes[
```
klemen1999: What does this resolves to in RVC2 (or RVC3)? We always add the U8 flag so should be uint8 right?

modelconverter/utils/types.py
```diff
@@ -203,6 +203,13 @@ def as_openvino_dtype(self) -> str:
     def as_snpe_dtype(self) -> str:
         return self.value
 
+    def as_nn_archive_dtype(self) -> str:
+        if self.value.startswith("ufxp"):
+            return self.value.replace("ufxp", "uint")
```
klemen1999, line 208: Should this be `uint8` and `int8` (added 8)?

kozlov721: It keeps the bitwidth suffix, just renames e.g. `ufxp8` to `uint8`.

modelconverter/utils/nn_archive.py
kozlov721: Fixed (well now it defaults to FP32 in case that's the correct value)

klemen1999 (approved):

modelconverter/utils/nn_archive.py
kozlov721: This turned out to be more tricky because users can theoretically override the `compile_tool` arguments and add custom `-ip`/`-op`.`-iop`, but it should be all handled correctly now.

PR #122, opened 2025-06-19
ONNX Split node compatibility

ptoupas:
## Purpose
During the creation of a custom [Split](https://onnx.ai/onnx/operators/onnx__Split.html) onnx node the newer ONNX versions (>=1.18.0) require `either input ‘split’ or the attribute ‘num_outputs’ should be specified`. Since we are adding such node on the modified ONNX model graph, we need to make sure the respective change is made during the creation of the Split node.

## Specification
Based on the opset version we add the split as input or attribute:
- opset < 13: split is added as an attribute
- opset >= 13: split is added as an input

Affects the modification of all the ONNX models during the conversion process.

## Deployment Plan
None / not applicable

## Testing & Validation
None / not applicable


klemen1999 (approved):
LGTM

PR #131, opened 2025-07-18
CI Updates

kozlov721:
## Purpose

Various CI updates.

## Specification

- Removed tests for RVC2 `v2021.4.0` because they're failing due to strange GH Actions-specific behavior of the host
  - `compile_tool` fails with an undefined symbol error, but it only ever happens when run in GH Actions 
- Updated pre-commit hooks to latest versions
- Removed old branch triggers

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/.github/workflows/unittests.yaml
+++ b/.github/workflows/unittests.yaml
@@ -2,7 +2,7 @@ name: Unit Tests
 
 on:
   pull_request:
-    branches: [dev]
+    branches: [main]
     paths:
       - requirements.txt
       - 'tests/test_utils/**'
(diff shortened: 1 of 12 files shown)


.github/workflows/unittests.yaml
```diff
@@ -2,9 +2,7 @@ name: Unit Tests
 
 on:
   pull_request:
-    branches:
-      - dev
-      - main
+    branches: [dev]
```
klemen1999: Why are these teid to `dev` branch?

kozlov721: That should be `main`, thanks

klemen1999 (approved):

PR #139, opened 2025-07-29
Fix a conversion error when a certain optimization applied before an Erf operation.

ptoupas:
## Purpose

- When the substitute_node_by_type (Div -> Mul) optimization is applied and an `Erf` op follows that creates an issue during the conversion of the onnx to dlc at the `snpe-onnx-to-dlc` command. This issue is not captured and does not affect the model at the onnx level.
- Sort the calibration images by name, which is necessary when working with multi-input modes, to keep all the inputs aligned.
- Check if the directory exists before deleting the raw images (some edge cases tries to delete the folder that does not exist producing an error)

## Specification
None / not applicable

None / not applicable

## Deployment Plan
None / not applicable

## Testing & Validation
None / not applicable

### DIFF
--- a/modelconverter/utils/onnx_tools.py
+++ b/modelconverter/utils/onnx_tools.py
@@ -554,6 +554,16 @@ class ONNXModifier:
 
         for node in self.onnx_gs.nodes:
             if node.op == source_node:
+                next_nodes = [
+                    n
+                    for n in self.onnx_gs.nodes
+                    if node.outputs[0] in n.inputs
+                ]
+                if any(n.op == "Erf" for n in next_nodes):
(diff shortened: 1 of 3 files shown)


modelconverter/utils/onnx_tools.py
```diff
@@ -554,6 +554,16 @@ def create_new_node(
 
         for node in self.onnx_gs.nodes:
             if node.op == source_node:
+                next_nodes = [
```
klemen1999, line 557: Let's add a comment here on which snpe version was this observed so we know to double check again at a later date if is still needed

klemen1999 (approved):

PR #143, opened 2025-07-31
Feature: Progress bars for docker pull

kozlov721:
## Purpose

Progress bars for a better user experience.

## Specification

Adds progress bars when pulling an image. 

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/modelconverter/utils/docker_utils.py
+++ b/modelconverter/utils/docker_utils.py
@@ -112,6 +112,8 @@ def docker_build(
     return image
 
 
+# We cannot simply call `docker pull` in a subprocess because
+# it interactively asks for login credentials if the image is private.
 def pull_image(client: docker.DockerClient, image: str) -> str:
     repository, tag = parse_repository_tag(image)
 
--- a/modelconverter/utils/docker_utils.py
+++ b/modelconverter/utils/docker_utils.py
@@ -128,7 +128,7 @@ def get_docker_image(
         docker_image = cast(Image, client.images.pull(f"ghcr.io/{image}", tag))
         docker_image.tag(image, tag)
 
-    except (docker.errors.APIError, docker.errors.DockerException):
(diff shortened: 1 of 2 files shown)


modelconverter/utils/docker_utils.py
```diff
@@ -104,6 +112,39 @@ def docker_build(
...
 
+# We cannot simply call `docker pull` in a subprocess because
+# it interactively asks for login credentials if the image is private.
+def pull_image(client: docker.DockerClient, image: str) -> str:
+    repository, tag = parse_repository_tag(image)
+
+    with Progress(
+        TextColumn("[progress.description]{task.description}"),
+        BarColumn(),
+        TaskProgressColumn(),
```
klemen1999, line 123: What if rich logging is disabled?

kozlov721: Hm, it should be alright since this gets executed before `modelconverter` actually starts (in the container) so it's not really part of the logs. But we can also disable it if rich logging is disabled.

klemen1999: Fair point yeah. And `rich` is already a dependency of modelconv from its dependencies right? In that case I think its also fine to keep, as you said: It shouldn't "spam" the actual logs

klemen1999 (approved):

modelconverter/utils/docker_utils.py
kozlov721: Actually, this gets executed even before we load the config, so we don't even know the status of rich logging yet

kozlov721: > And rich is already a dependency of modelconv from its dependencies right? 

yeah, it's a dependency of `luxonis-ml`

PR #145, opened 2025-08-01
RVC4: Simpler support for setting the optimization level.

ptoupas:
## Purpose
Makes it easier to set the optimization level used for the DLC graph preparation on RVC4.

## Specification
- Added new configuration field `rvc4.optimization_level` which will set the respective flag to snpe-dlc-graph-prepare as regards the optimization level used for the DLC graph preparation.

None / not applicable

## Deployment Plan
None / not applicable

## Testing & Validation
None / not applicable

### DIFF
--- a/tests/test_utils/test_config.py
+++ b/tests/test_utils/test_config.py
@@ -65,6 +65,7 @@ DEFAULT_TARGET_CONFIGS = {
         "use_per_channel_quantization": True,
         "use_per_row_quantization": False,
         "compress_to_fp16": False,
+        "optimization_level": 2,
     },
     "hailo": {
         "optimization_level": 2,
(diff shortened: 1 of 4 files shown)


tests/test_utils/test_config.py
```diff
@@ -65,6 +65,7 @@
         "use_per_channel_quantization": True,
         "use_per_row_quantization": False,
         "compress_to_fp16": False,
+        "optimization_level": 2,
```
klemen1999, line 68: What happens if this is specified through this special flag and also through the snpe_dlc_graph_prepare_args? Right now it will get then set twice in the underlying snpe command. Does this cause issues or is it just ignored by snpe?

kozlov721: This is handled by the `BaseExporter._add_args` method. Args specified in `snpe_dlc_graph_prepare_args` take precedence over anything.

kozlov721 (approved):

tests/test_utils/test_config.py
klemen1999: Oh I see [this](https://github.com/luxonis/modelconverter/blob/c281c77d61643a8e9c59277cbcf4a7029c93a1db/modelconverter/packages/base_exporter.py#L232) extra check in the _add_args() now yeah, ty.

klemen1999 (approved):

PR #149, opened 2025-08-12
Fix: Expand analyzer input types and add logger

aljazkonec1:
## Purpose

This PR expands analyzer to accept npy input types, adds a logger and fixes issues regarding output name checking and comparison.

## Specification

Changes: 
- Add logger to analyze tool
- Option to specify adb device id instead of the tool automatically picking the first one.
- add support for `.npy` input file types ( tool assumes all preprocessing has been applied and that the npy files are in FP32 format)
- Instead of a hardcoded single transformation option from NCHW -> NHWC, added all possible transformations from 2 dims to 5 dims ( `transpose_to_match` method)
- added `mean_absolute_percentage_error` metric, round all metrics to 6 decimal places
- Improved layer name sanitation due to problems with `strip` method.
-
## Deployment Plan

None / not applicable

## Testing & Validation

Tested with YOLOE model and some images.

### DIFF
--- a/modelconverter/__main__.py
+++ b/modelconverter/__main__.py
@@ -369,9 +369,12 @@ def benchmark(
 @app.meta.command(group=device_commands)
 def analyze(
     *,
+    device_id: str | None = None,
     dlc_model_path: str,
     onnx_model_path: str,
-    image_dirs: Annotated[list[str], Parameter(negative_iterable=[], consume_multiple=True)],
+    image_dirs: Annotated[
+        list[str], Parameter(negative_iterable=[], consume_multiple=True)
+    ],
     analyze_outputs: bool = True,
     analyze_cycles: bool = True,
 ) -> None:
@@ -412,7 +415,9 @@ def analyze(
                 for i in range(0, len(image_dirs), 2)
             }
 
-        analyzer = get_analyzer(Target.RVC4, dlc_model_path, image_dirs_dict)
+        analyzer = get_analyzer(
+            Target.RVC4, device_id, dlc_model_path, image_dirs_dict
+        )
         if analyze_outputs:
             logger.info("Analyzing layer outputs")
             analyzer.analyze_layer_outputs(
--- a/modelconverter/__main__.py
+++ b/modelconverter/__main__.py
@@ -365,7 +365,6 @@ def benchmark(
     get_benchmark(target, model_path).run(full=full, save=save, **kwargs)
 
 
-# TODO: Specify device ID in case more than one device is connected
 @app.meta.command(group=device_commands)
 def analyze(
     *,
@@ -385,6 +384,8 @@ def analyze(
 
     Parameters
     ----------
+    device_id : str | None
+        The ID of the device to run the analysis on. If not provided, the first listed device will be used.
     dlc_model_path : str
         The path to the DLC model file.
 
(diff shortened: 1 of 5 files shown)


klemen1999 (approved):

klemen1999:
CC: @ptoupas if you can take a look as well

kozlov721 (approved):
LGTM

modelconverter/__main__.py
```diff
@@ -369,9 +369,12 @@ def benchmark(
 @app.meta.command(group=device_commands)
 def analyze(
     *,
+    device_id: str | None = None,
```
ptoupas, line 371: Could we add the respective part in the docstring below for that parameter?
Also I think we should remove the `# TODO` comment above (line 368).

modelconverter/packages/rvc4/analyze.py
```diff
@@ -75,23 +84,30 @@ def _prepare_raw_inputs(
             input_row = ""
             dlc_matcher[i] = "Result_" + str(i)
             for input_name, img_path in input_dict.items():
-                if not img_path.endswith((".png", ".jpg")):
+                if not img_path.endswith((".png", ".jpg", ".npy")):
```
ptoupas: I suppose we could add the `.jpeg` suffix as well here.

modelconverter/packages/rvc4/analyze.py
```diff
@@ -144,28 +192,35 @@ def _compare_to_onnx(
...
         onnx_input_shapes = {}
+        logger.info("Inferencing ONNX model.")
         for input_metadata in session.get_inputs():
             onnx_input_shapes[input_metadata.name] = input_metadata.shape
 
         for i, input_dict in input_matcher.items():
             onnx_input_dict = {}
             for input_name, img_path in input_dict.items():
-                if not img_path.endswith((".png", ".jpg")):
+                if not img_path.endswith((".png", ".jpg", ".npy")):
```
ptoupas: Same as above, we could add the `.jpeg` suffix as well.

modelconverter/packages/rvc4/analyze.py
```diff
@@ -75,23 +84,30 @@ def _prepare_raw_inputs(
...
-                width_height = self.input_sizes[input_name][1:3][::-1]
-                image = self._resize_image(img_path, width_height)
-                image = image.astype(type)
-                raw_image = cast(np.ndarray, image).astype(type)
+                if img_path.endswith(
+                    ".npy"
+                ):  # expects numpy array to already be in correct format
+                    raw_image = cast(np.ndarray, np.load(img_path)).astype(
+                        type
+                    )
```
ptoupas, line 96: Even though we expect here the .npy file to be at the correct format, I still think it would be good to have a check on the numpy array shape to ensure its aligned with the expected shape.

modelconverter/packages/rvc4/analyze.py
```diff
@@ -144,28 +192,35 @@ def _compare_to_onnx(
...
-                shape = onnx_input_shapes[input_name][2:][::-1]
-                image = self._resize_image(img_path, shape)
-                image = np.transpose(
-                    image, [2, 0, 1]
-                )  # NCHW format is assumed by default and resize returns HWC
-                image = np.expand_dims(image, axis=0).astype(np.float32)
+                if img_path.endswith(
+                    ".npy"
+                ):  # expects numpy array to already be in correct format
+                    image = np.load(img_path)
```
ptoupas, line 212: Same as above, lets add a check on the shape of the numpy array.

ptoupas (changes requested):
I've left some comments to be addressed, other than that it LGTM.

ptoupas (approved):

PR #155, opened 2025-08-21
Fix DLC metadata parsing on edge cases.

ptoupas:
## Purpose
Apply an enhancement/fix on the parsing of the dlc metadata. There are some edge cases where apart from the Inputs and Outputs there are also the Unconsumed Tensors in the graph that are reported in the output of the `snpe-dlc-info` command.
This PR takes that into account when parsing and looking for the input/output dimensions, types and encodings addressing an error occurs with the generation of the final NNArchive.

## Specification
The PR deals with outputs of the `snpe-dlc-info` command that look like the one below:
```bash
------------------------------------------------------------------------------
| Input Name  | Dimensions   | Type      | Encoding Info                     |
------------------------------------------------------------------------------
| input       | 1,480,640,3  | Float_16  | No encoding info for this tensor  |
------------------------------------------------------------------------------
------------------------------------------------------------------------------
| Output Name  | Dimensions  | Type      | Encoding Info                     |
------------------------------------------------------------------------------
| output       | 100,6       | Float_16  | No encoding info for this tensor  |
------------------------------------------------------------------------------
------------------------------------------------------------------------------------------------------------------------
| Unconsumed Tensor Name                                  | Dimensions  | Type     | Encoding Info                     |
------------------------------------------------------------------------------------------------------------------------
| /NonMaxSuppression_output_0_valid_num_selected_indices  | 1           | Uint_32  | No encoding info for this tensor  |
------------------------------------------------------------------------------------------------------------------------
Total parameters: ...
```

None / not applicable

## Deployment Plan
None / not applicable

## Testing & Validation
None / not applicable

### DIFF
--- a/modelconverter/utils/metadata.py
+++ b/modelconverter/utils/metadata.py
@@ -54,16 +54,60 @@ def _get_metadata_dlc(path: Path) -> Metadata:
     metadata = {}
 
     for typ in ["input", "output"]:
-        start_marker = f"{typ.capitalize()} Name,Dimensions,Type,Encoding Info"
-        if typ == "input":
-            end_marker = "Output Name,Dimensions,Type,Encoding Info"
-        else:
-            end_marker = "Total parameters:"
-        start_index = content.find(start_marker)
-        end_index = content.find(end_marker, start_index)
+        header_pattern = f"{typ.capitalize()} Name"
 
-        relevant_csv_part = content[start_index:end_index].strip()
-        if not relevant_csv_part:
+        start_index = content.find(header_pattern)
+        if start_index == -1:
+            continue
+
+        line_start = content.rfind("\n", 0, start_index) + 1
+        possible_endings = []
+
+        if typ == "input":
+            output_idx = content.find(
+                "Output Name", start_index + len(header_pattern)
+            )
+            if output_idx != -1:
+                possible_endings.append(output_idx)
+        else:
+            unconsumed_idx = content.find(
+                "Unconsumed Tensor Name", start_index + len(header_pattern)
+            )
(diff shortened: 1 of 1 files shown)


modelconverter/utils/metadata.py
```diff
@@ -54,16 +54,60 @@ def _get_metadata_dlc(path: Path) -> Metadata:
...
         if typ == "input":
-            end_marker = "Output Name,Dimensions,Type,Encoding Info"
+            output_idx = content.find(
+                "Output Name", start_index + len(header_pattern)
+            )
+            if output_idx != -1:
+                possible_endings.append(output_idx)
+        else:
+            unconsumed_idx = content.find(
+                "Unconsumed Tensor Name", start_index + len(header_pattern)
```
klemen1999, line 74: The unconsumed tensors get ignored right?

modelconverter/utils/metadata.py
klemen1999, line 1: Is the output format dependant on the SNPE version used? Can we check that this improved parsing works with few different snpe versions?

klemen1999 (commented):
Unconsumm

modelconverter/utils/metadata.py
ptoupas: Yeah, that's the fix actually to ignore the `Unconsumed Tensor` rows

modelconverter/utils/metadata.py
ptoupas: I've tested that change on two different models both using SNPE 2.23.0 and SNPE 2.32.6 and it seems to work fine in each possible combination.

klemen1999 (approved):

PR #159, opened 2025-08-28
Small fixes on onnx tool and rvc2 exporter

ptoupas:
## Purpose
Apply small fixes and fix compatibility issues (with ONNX >=1.19.0)  on `onnx tool`, `rvc2 exporter`,, and `rvc4 benchmark`.

## Specification
None / not applicable

None / not applicable

## Deployment Plan
None / not applicable

## Testing & Validation
None / not applicable

### DIFF
--- a/modelconverter/packages/rvc2/exporter.py
+++ b/modelconverter/packages/rvc2/exporter.py
@@ -71,7 +71,7 @@ class RVC2Exporter(Exporter):
                     inp_str += ","
                 inp_str += name
                 if inp.shape is not None:
-                    inp_str += f"{_lst_join(inp.shape, sep=',')}"
+                    inp_str += f"{_lst_join(inp.shape, sep=' ')}"
                 if inp.data_type is not None:
                     if OV_2021 and self.compress_to_fp16:
                         data_type = DataType("float16")
--- a/modelconverter/packages/rvc2/exporter.py
+++ b/modelconverter/packages/rvc2/exporter.py
@@ -71,7 +71,7 @@ class RVC2Exporter(Exporter):
                     inp_str += ","
                 inp_str += name
                 if inp.shape is not None:
-                    inp_str += f"{_lst_join(inp.shape, sep=' ')}"
+                    inp_str += f"{_lst_join(inp.shape, sep=',')}"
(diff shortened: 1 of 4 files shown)


modelconverter/packages/rvc2/exporter.py
```diff
@@ -71,7 +71,7 @@ def _export_openvino_ir(self) -> Path:
                     inp_str += ","
                 inp_str += name
                 if inp.shape is not None:
-                    inp_str += f"{_lst_join(inp.shape, sep=' ')}"
+                    inp_str += f"{_lst_join(inp.shape, sep=',')}"
```
kozlov721: I think this will break on the `2021.4` version of openvino. This version is atm not tested automatically because of strange issues in the GH Actions environment. Could you test it locally to see if it still works?

kozlov721: The same goes for similar changes below

requirements.txt
```diff
@@ -1,7 +1,7 @@
 Pillow
 gcsfs
 luxonis-ml[data,nn_archive]==0.8.0
-onnx==1.17.0
+onnx>=1.17.0,<1.19.0
```
kozlov721, line 4: Do we still need this with the `mapping` fix above?

ptoupas: We could keep just the `1.17.0` or allow for both `1.17.0` and `1.18.0`. But we cannot allow the `1.19.0` since that gives some errors. @kozlov721, is there a reason we DON'T want the `1.18.0` version?

modelconverter/packages/rvc2/exporter.py
ptoupas: I've tested the `Lite-HRNet - 18 COCO 192x256` model both with `2022.3.0` and `2021.4.0` and the addition of the `,` does indeed give an error at `2021.4.0` version. I will revert that change back.
The changes on the mean and scale values below do not affect any of the versions.

requirements.txt
kozlov721: I don't think there's a reason we don't want `1.18.0`. If it works we should use the latest version possible

kozlov721 (approved):

PR #168, opened 2025-09-23
Handle manual FP16 compression flags when setting NNArchive precision

ptoupas:
## Purpose
NNArchive precision was only set to FP16 when compress_to_fp16 was used. Since in RVC4 the same can be achieved with manual flags (`--float_bitwidth 16` and `--use_float_io`), update logic to recognize these as well to ensure consistent FP16 precision in the output archive.

## Specification

- Added detection of "--float_bitwidth" "16" pair in snpe_onnx_to_dlc_args.
- Checked for "--use_float_io" in snpe_dlc_graph_prepare_args.
- Updated compress_to_fp16 to be true if either the explicit flag or manual args are set.

None / not applicable

## Deployment Plan
None / not applicable

## Testing & Validation
None / not applicable

### DIFF
--- a/modelconverter/utils/nn_archive.py
+++ b/modelconverter/utils/nn_archive.py
@@ -1,5 +1,6 @@
 import json
 import tarfile
+from itertools import pairwise
 from pathlib import Path
 from typing import Any, Literal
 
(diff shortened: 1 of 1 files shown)


kozlov721 (commented):
This could be read from the final DLC (or the `info.csv` file we generate during the conversion) which would be more reliable way to determine this. We have the `get_metadata` function in `utils.metadata` for this.

ptoupas:
> This could be read from the final DLC (or the `info.csv` file we generate during the conversion) which would be more reliable way to determine this. We have the `get_metadata` function in `utils.metadata` for this.

@kozlov721 
That would give us the dtypes for each input and output of the model but in multi-input models with different dtypes for different inputs we would't know which input's dtype to choose.
We can always pick the first input from the `input_dtypes` dictionary of the metadata but there might be a case where this won't be correct (edge case though).

Also I am not quite sure what is the suggestion here. Do you propose to completely remove the `match... case` block and just infer the `precision` from the `model_metadata = get_metadata(model_path)` instead?

kozlov721 (approved):

kozlov721:
Yeah I see, for now this will be the easiest solution. My suggestion would be for RVC4 to read the inputs and decide based on their precision such that the "lowest" precision beats the higher one (for models with more inputs), but that would require some more verifying whether it works correctly.

ptoupas:
> Yeah I see, for now this will be the easiest solution. My suggestion would be for RVC4 to read the inputs and decide based on their precision such that the "lowest" precision beats the higher one (for models with more inputs), but that would require some more verifying whether it works correctly.

So, should we merge this PR now and create another task to deal with the multi-input models using the `get_metadata`?

kozlov721:
>
So, should we merge this PR now and create another task to deal with the multi-input models using the get_metadata?

Yeah

PR #179, opened 2025-11-11
Feat: `quantization_mode`

ptoupas:
## Purpose
Introduce a new quantization related feature to the modelconverter tool by adding an optional, high-level configuration parameter (`quantization_mode`) to simplify model quantization during RVC4 convertion.
The goal is to make quantization strategies more accessible through the UI, while still allowing full control through advanced flags.

## Specification
- Introduced a new `quantization_mode` parameter for RVC4 model conversion, which has 5 available options: [`INT8_STANDARD`, `INT8_ACCURACY_FOCUSED`, `INT8_INT16_MIXED`, `FP16_STANDARD`, `CUSTOM`]. Defaults to `INT8_STANDARD` and the current default behavior of the quantized conversion should not be affected.
- When this parameter is set to anything but `CUSTOM` it will take precedence over any user-defined SNPE flags provided via the `snpe_onnx_to_dlc_args`, `snpe_dlc_quant_args`, and `snpe_dlc_graph_prepare_args`.
- Removed the `compress_to_fp16` parameter which is now replaced by the `FP16_STANDARD` mode in the `quantization_mode` parameter.
- When the `quantization_mode` is set to `INT8_INT16_MIXED` a .json encodings file is automatically generated to keep the inputs and outputs of the model into 8-bit precision so that the generated DLC can be executed and be compatible with DAI.
- The remaining RVC4 related parameters such as `use_per_channel_quantization`, `use_per_row_quantization`, `optimization_level`, and others, should not be affected by the changes introduced in that PR.

None / not applicable

## Deployment Plan
None / not applicable

## Testing & Validation
Tested the conversion of the following models under different configurations and settings through the CLI.
- [x] [ConvNeXt](https://hub.luxonis.com/ai/models/91311c73-e8c8-414e-8a57-41bf944fa71c?view=page)
- [x] [YOLOv6 Nano](https://hub.luxonis.com/ai/models/face58c4-45ab-42a0-bafc-19f9fee8a034?view=pagehttps://hub.luxonis.com/ai/models/face58c4-45ab-42a0-bafc-19f9fee8a034?view=page)
- [x] [YuNet](https://hub.luxonis.com/ai/models/5d635f3c-45c0-41d2-8800-7ca3681b1915?view=page)
- [x] [Paddle Text Recognition](https://hub.luxonis.com/ai/models/9ae12b58-3551-49b1-af22-721ba4bcf269?view=page)
- [x] ResNet18 (local model)
- [x] SMv4 (local model)

### DIFF
--- a/modelconverter/hub/README.md
+++ b/modelconverter/hub/README.md
@@ -133,7 +133,6 @@ Parameters specific to the `RVC4` conversion.
 | `snpe_dlc_graph_prepare_args`  | `list[str] \| None` | The arguments to pass to the `snpe-dlc-graph-prepare` tool.  |
 | `use_per_channel_quantization` | `bool`              | Whether to use per-channel quantization. Defaults to `True`. |
 | `use_per_row_quantization`     | `bool`              | Whether to use per-row quantization. Defaults to `False`.    |
-| `quantization_mode`            | `Literal["INT8_STANDARD", "INT8_ACCURACY_FOCUSED", "INT8_INT16_MIXED", "FP16_STANDARD", "CUSTOM"]`  | The pre-defined quantization mode to use. Defaults to `INT8_STANDARD`.   |
 | `htp_socs`                     | `list[str] \| None` | The list of HTP SoCs to use.                                 |
 
 **Hailo Parameters**
--- a/modelconverter/hub/README.md
+++ b/modelconverter/hub/README.md
@@ -133,6 +133,7 @@ Parameters specific to the `RVC4` conversion.
 | `snpe_dlc_graph_prepare_args`  | `list[str] \| None` | The arguments to pass to the `snpe-dlc-graph-prepare` tool.  |
 | `use_per_channel_quantization` | `bool`              | Whether to use per-channel quantization. Defaults to `True`. |
 | `use_per_row_quantization`     | `bool`              | Whether to use per-row quantization. Defaults to `False`.    |
+| `quantization_mode`            | `Literal["INT8_STANDARD", "INT8_ACCURACY_FOCUSED", "INT8_INT16_MIXED", "FP16_STANDARD", "CUSTOM"]`  | The pre-defined quantization mode to use. Defaults to `INT8_STANDARD`.   |
 | `htp_socs`                     | `list[str] \| None` | The list of HTP SoCs to use.                                 |
 
 **Hailo Parameters**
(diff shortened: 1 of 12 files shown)


modelconverter/hub/README.md
klemen1999, line 1: Changes inside `hub/` are not needed since this will be removed. 
CC: @kkeroo on the PR for visibility - once merged in and exposed in the API through Hub we would want to update the SDK accordingly

ptoupas: Ok, I will revert these changes in that case. Thanks for letting me know about that.

modelconverter/packages/rvc4/exporter.py
```diff
@@ -46,6 +47,14 @@ def __init__(self, config: SingleStageConfig, output_dir: Path):
         )
         self.use_per_row_quantization = rvc4_cfg.use_per_row_quantization
         self.optimization_level = rvc4_cfg.optimization_level
+        self.quantization_mode = rvc4_cfg.quantization_mode
+        if self.quantization_mode != QuantizationMode.CUSTOM:
+            self.snpe_onnx_to_dlc = []
+            self.snpe_dlc_quant = []
+            self.snpe_dlc_graph_prepare = []
+            logger.warning(
+                f"Overriding user-provided SNPE arguments. Using pre-defined arguments for quantization mode {self.quantization_mode.value}."
```
klemen1999: I'd reword this a bit to tell what can they do if this is not desired. Maybe something like this:
"Using pre-defined arguments for quantization mode {self.quantization_mode.value} which overrides user-provided SNPE arguments. If you want to manually set SNPE arguments set `rvc4.quantization_mode: CUSTOM` in config."

modelconverter/utils/types.py
```diff
@@ -229,6 +229,15 @@ class Target(Enum):
     RVC4 = "rvc4"
 
 
+class QuantizationMode(Enum):
+    INT8_STD = "INT8_STANDARD"
+    INT8_ACC = "INT8_ACCURACY_FOCUSED"
+    INT8_16_MIX = "INT8_INT16_MIXED"
+    INT8_FP16_MIX = "INT8_FP16_MIXED"  # Added for future use
```
klemen1999: Should we raise some "Not implemented" error if this is set? Or should we maybe comment it out completely for now?

klemen1999 (approved):
Left some minor comments but generally LGTM.

modelconverter/utils/types.py
ptoupas: Yeah, I completely removed it for now, once we sort out the issue with encodings with QC we can add this one back.

PR #185, opened 2025-11-28
feat: add ADB monitoring

jkbmrz:
## Purpose

This change adds support for collecting additional performance metrics—power consumption and DSP utilization. Benchmarking is performed via ADB; therefore, this functionality is limited to RVC4.

## Specification

The modelconverter benchmark CLI now includes:
- `power_benchmark` and `dsp_benchmark` options
- Extended `BenchmarkResult` structure to store power and DSP metrics
- Updated `Benchmark.print_results()` to handle extended result rows
- New ADB-based monitors: `AdbMonitorPower` and `AdbMonitorDSP`
- Utility function `get_adb_id` for resolving device ADB identifiers

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

Tested using `luxonis/yunet:640x480` with default benchmarking mode.


jkbmrz:
Open Issues:

**1. Idle measurements**
Should we also record and report idle power/DSP measurements? If yes, should idle be:
- reported as an additional standalone metric, or
- automatically subtracted from the measured values to provide a “net” load value?

**2. `device_id` vs. `device_ip` ambiguity**
Our `AdbHandler` requires device_id which is different from device_ip (even if they both refer to the same physical camera). M is not straightforward. Ideally, the user would only need to provide the IP, but right now there is no clean, universal way to map from one to another.

I introduced `get_adb_id()` as a helper, which works in the current test environment where the ADB ID matches the hex-encoded MXID. However, this is likely not guaranteed across all devices or setups. Does anyone know of a more robust way to map IP → adb_id across devices? If not, we might need to require the user to explicitly provide the adb_id when enabling power/DSP benchmarking.

ptoupas:
Added a few extra features for power and dsp utilization benchmarking:

- Automatically calculate the idle power and dsp utilization before the benchmark starts, and use these values to filter the metrics lists during benchmarking.
- Add the option for the user to set the device by either providing the `device_ip` or the `device_mxid`. Internally we have introduced a way to map either of these into the respective ADB ID which is needed to run ADB commands for both power and dsp utilization.
- For DSP utilization, we do not longer require from the user to setup the necessary files into the target device. We instead create them before the benchmark starts and delete them once its done.

modelconverter/packages/rvc4/benchmark.py
```diff
@@ -69,6 +72,7 @@ def default_configuration(self) -> Configuration:
             "num_threads": 2,
             "num_messages": 50,
             "device_ip": None,
+            "device_mxid": None,
```
klemen1999: Let's rename `device_mxid` -> `device_id` and generally use `deviceId` instead of `mxId`

klemen1999 (commented):
Code wise LGTM, I'll test it out tomorrow. Thanks

modelconverter/utils/adb_monitor.py
```diff
@@ -177,15 +181,28 @@ def get_stats(self) -> tuple[float | None, float | None]:
             `(mean_power0, mean_power1)` in watts. Each value is None if no
             valid samples were collected for that channel.
         """
-        # mypy/pyright: measurements are Sequence[object]; we know actual type here
         samples = [m for m in self.measurements if isinstance(m, tuple) and len(m) == 2]
-        p0_vals = [p0 for (p0, _) in samples if p0 is not None]
-        p1_vals = [p1 for (_, p1) in samples if p1 is not None]
+        p0_vals = [p0 for (p0, _) in samples if p0 is not None and p0 > self.idle_power_system]
+        p1_vals = [p1 for (_, p1) in samples if p1 is not None and p1 > self.idle_power_processor]
```
jkbmrz: This is something I was wondering about before - why do we need this `p > idle_power` check?

modelconverter/utils/adb_monitor.py
```diff
@@ -177,15 +181,28 @@ def get_stats(self) -> tuple[float | None, float | None]:
...
 
         mean_p0 = statistics.fmean(p0_vals) if p0_vals else None
         mean_p1 = statistics.fmean(p1_vals) if p1_vals else None
         return mean_p0, mean_p1
 
+    def set_idle_power(self) -> None:
+        logger.info("Calculating idle power consumption...")
+        self.start()
+        time.sleep(5)
```
jkbmrz: Should we set a bit longer time period here? Using the 0.5s sampling interval, we make 10 measurements which might not be enough to determine the average well in cases when idle power is not that stable.

modelconverter/utils/adb_monitor.py
```diff
@@ -0,0 +1,403 @@
...
+            Name of the hwmon node (e.g. "hwmon0", "hwmon1").
+
+        Returns
+        -------
+        bool
+            True if the file exists, False otherwise.
+        """
+        try:
+            self.adb_handler.shell(f"ls /sys/class/hwmon/{hwmon}/power1_input")
+            time.sleep(self.interval)
```
jkbmrz: Why do we need `time.sleep` here? Same question for `_check_dsp` method (line 327)

modelconverter/utils/adb_monitor.py
```diff
@@ -0,0 +1,403 @@
...
+        ]
+
+        mean_p0 = statistics.fmean(p0_vals) if p0_vals else None
+        mean_p1 = statistics.fmean(p1_vals) if p1_vals else None
+        return mean_p0, mean_p1
+
+    def set_idle_power(self) -> None:
+        logger.info("Calculating idle power consumption...")
+        self.start()
+        time.sleep(5)
```
jkbmrz, line 276: Should we set a bit longer time period here? Using the 0.5s sampling interval, we make 10 measurements which might not be enough to determine the average well in cases when idle power is not that stable.

modelconverter/utils/adb_monitor.py
```diff
@@ -0,0 +1,403 @@
...
+        if self.dsp_exists:
+            self.idle_dsp_utilization: float = 0.0
+            self.set_idle_dsp()
+
+    def _prepare_dsp_util_script(self) -> None:
+        """Create the DSP utility script directly on the device via
+        ADB."""
+        remote_script_path = "/data/local/oak_dsp_util.sh"
+
+        script_content = r"""SYS_MON_APP="/usr/bin/sysMonApp"
```
jkbmrz: Should this go somewhere else for better code readability? E.g. in a separate file or defined as a constant outside of the class?

modelconverter/utils/adb_monitor.py
```diff
@@ -0,0 +1,403 @@
...
+        """
+        samples = [
+            m
+            for m in self.measurements
+            if isinstance(m, tuple) and len(m) == 2
+        ]
+        p0_vals = [
+            p0
+            for (p0, _) in samples
+            if p0 is not None and p0 > self.idle_power_system
```
jkbmrz, line 261: This is something I was wondering about before - why do we need this `p > idle_power` check? If a process we measure consumes almost no power, we will report "N/A" as if the power couldn't be measured (while it's just too close to the idle value).

modelconverter/utils/adb_monitor.py
ptoupas: We need this to prevent this warning to be displayed moving from the first time we check for the hwmon or DSP files to the first actual measurement.
`WARNING  Failed to read hwmon0 power value.`

modelconverter/utils/adb_monitor.py
ptoupas: From my experience so far the idle power is quite stable and just a few measurements should be fine, but let me know if you have experienced big differences and unstable behaviour with the idle power measurements, and if so I can increase this time a bit.
I just don't want to do it without a good reason, because the benchmarking will take longer and there is no reason for the user to wait that long.

modelconverter/utils/adb_monitor.py
ptoupas: I don't thing there is any case where a NN model will consume power that is close to the idle case. It should always be much higher than the idle power.
The reason we do this check is because the list of power measurements sometimes has measurements that are a bit behind in time (shifted) so the device is still idle, that's why we need to filter those values to get more accurate average estimates.

modelconverter/utils/adb_monitor.py
ptoupas: Moved it to a constant at the start of the file.

modelconverter/utils/adb_monitor.py
ptoupas: Moved it to the `start()` function of the `_BaseAdbMonitor` instead.

modelconverter/packages/rvc4/analyze.py
```diff
@@ -13,20 +13,23 @@
...
         dlc_model_path: str,
         image_dirs: dict[str, str],
     ):
         super().__init__(dlc_model_path, image_dirs)
-        self.adb = AdbHandler(device_id)
+        _, device_adb_id = get_device_info(device_ip, device_id)
+        self.adb = AdbHandler(device_adb_id)
 
-        self.device_id = device_id
+        self.device_id = device_adb_id
```
jkbmrz: Shouldn't this be `self.device_adb_id`?

modelconverter/__main__.py
```diff
@@ -411,8 +424,10 @@ def analyze(
 
     Parameters
     ----------
+    device_ip : str | None
+        IP address of the device to run the benchmark on. Interchangeable with device_id. If neither is given, DAI selects the default device. If both are given, device_id takes precedence.
     device_id : str | None
-        The ID of the device to run the analysis on. If not provided, the first listed device will be used.
+        DeviceId of the device to run the benchmark on. Interchangeable with device_ip. If neither is given, DAI selects the default device. If both are given, device_id takes precedence.
```
jkbmrz, line 430: `DeviceId of the device to run the benchmark on` is pretty undescriptive. Could we explain better what's the ID we expect here?

modelconverter/utils/adb_monitor.py
jkbmrz: I see, that makes sense! Should we just perhaps use a different sleep duration instead of `self.interval`? While this works, reusing the sampling interval here feels a bit misleading — it wasn’t immediately clear to me why it’s tied to the value representing the sampling interval. Maybe we could use a fixed duration (e.g., 1s) or something that better reflects the time needed for the hwmon/DSP check to complete?

modelconverter/utils/adb_monitor.py
jkbmrz: I haven't tested the stability of idle values myself but I believe you this is sufficient!

jkbmrz (commented):
Did another check — LGTM now!
Just one small comment regarding the `device_id` / `device_ip` / `device_adb_id` naming. I think it would help if we better clarify what each of these represents and where each should be used. We previously used `device_id` to refer to what is now `device_adb_id`, which could cause some confusion for users.

modelconverter/packages/rvc4/analyze.py
ptoupas: Actually this is never used, so I am completely removing it.

modelconverter/__main__.py
klemen1999: I think in the context of working with the device in DAIv3 it is clear enough what would `DeviceId` be. And if one wouldn't be sure they can look this up in the docs, I checked and the ChatBot refers you to [this](https://docs.luxonis.com/software-v3/depthai/examples/misc/device_information) page which should help you get this ID.

klemen1999 (approved):
LGTM

PR #186, opened 2025-12-05
Feat: RVC2 & RVC4 benchmarking use benchmark-time as default

ptoupas:
## Purpose
Changed the default way to perform benchmarking from repetition-based to time-based.

## Specification
Set the `benchmark-time` to 20 seconds by default (instead of None) which takes precedence over `repetitions`.
To use the repetition-based benchmark the user will need to set the `benchmark-time` to a negative value (`--benchmark-time -1`)

None / not applicable

## Deployment Plan
None / not applicable

## Testing & Validation
Tested the default flags and setting the `benchmark-time` to a negative value during `yolov6-nano` benchmarking

### DIFF
--- a/modelconverter/__main__.py
+++ b/modelconverter/__main__.py
@@ -306,9 +306,7 @@ def benchmark(
     full: bool = False,
     save: bool = False,
     repetitions: Annotated[int, Parameter(group=["RVC2", "RVC4"])] = 10,
-    benchmark_time: Annotated[
-        int | None, Parameter(group=["RVC2", "RVC4"])
-    ] = None,
+    benchmark_time: Annotated[int, Parameter(group=["RVC2", "RVC4"])] = 20,
     num_threads: Annotated[int, Parameter(group=["RVC2", "RVC4"])] = 2,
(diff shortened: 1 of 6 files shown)


kozlov721 (approved):

modelconverter/__main__.py
```diff
@@ -306,9 +306,7 @@ def benchmark(
     full: bool = False,
     save: bool = False,
     repetitions: Annotated[int, Parameter(group=["RVC2", "RVC4"])] = 10,
-    benchmark_time: Annotated[
```
klemen1999, line 309: Why the change to not acceprt None anymore? Does it cause problems with setting None through CLI?

ptoupas: Yeah there is no way to accept None from CLI, so the two options are:

1. Check the `benchmark_time` and when its negative consider it as "disabled". (the one I chose)
2. Create a new bool flag something like `disable_benchmark_time` and use that to actually disable the `benchmark_time` and use the `repetitions` instead.

klemen1999: Ok then I agree, the mehtod you choose makes more sense

klemen1999 (approved):

PR #192, opened 2025-12-15
Update README regarding local conversions

ptoupas:
## Purpose
Update README regarding local conversions and removed unused packages from requirements.

## Specification
None / not applicable

None / not applicable

## Deployment Plan
None / not applicable

## Testing & Validation
None / not applicable


klemen1999 (commented):
LGTM, let's update the [docs](https://docs.luxonis.com/) as well accordingly.

I'll leave the PR "open" in case we want to add some more additions to it

README.md
```diff
@@ -210,7 +216,7 @@ docker build -f docker/$TARGET/Dockerfile \
              -t luxonis/modelconverter-$TARGET:latest .
 ```
 
-If you want to build the image with a different version of the underlying conversion tools than is the default one, you also need to pass the `--build-arg` flag with the desired version. For example, to build the `RVC2` image with ` 2021.4.0`, use:
+If you want to build the image with a different version of the underlying conversion tools than is the default one, you also need to pass the `--build-arg` flag with the desired version. For example, to build the `RVC2` image with `2021.4.0`, use:
```
klemen1999, line 219: Given that we below write:

> If you build images using a generic `:latest` tag (for example `luxonis/modelconverter-rvc2:latest`), the CLI will not detect them and will instead attempt to pull the expected tag from the registry.

wouldn't it make sense to alread here (and the example above) not use just `:latest` tag?
Because if I follow the instructions now from start to finish I'll build the image with :latest tag and below this will not get picked up and the image I prepared won't get used

ptoupas: That's true, if the `:latest` tag is used the locally prepared image won't get used by default but that's why we have introduced the new `--docker-image` flag, so that even if the docker image does not comply with the default and expected naming conventions of the CLI tool, it can still be used via this flag.
But I get what you mean, we can update all the procedure to use the `:2.32.6-latest` tag instead so that the use of the  `--docker-image` flag is not needed.

klemen1999: Yep this is what I meant. Basically the flow we have explicitly in here should work as is, it should be a "happy path" not something where you need to add extra flags to make it work.
And putting 2.32.6-latest is maybe fine but then we need to put a note that this is for 2.32.6 version explicitly and if some other version is used then this needs to be replaced accordingly

ptoupas: I've made some changes to the README based on your comments. I believe things are more clear now, and when someone follows the flow and the commands provided as examples everything should be working.

klemen1999 (approved):
LGTM

kozlov721 (approved):

PR #194, opened 2025-12-16
Feat/add docker image tag option

kozlov721:
## Purpose

It makes it possible for users to pass a specific docker image to the CLI. 

## Specification

- Added new `--image/--docker-image` argument to the relevant CLI commands.
- If the image name includes a tag, it will be used as is
- In other cases a tag will be constructed as usually (`{version}-latest`) and used with the user provided image name 

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/modelconverter/__main__.py
+++ b/modelconverter/__main__.py
@@ -574,6 +574,14 @@ def launcher(
             "Available options differ based on the target platform. ",
         ),
     ] = None,
+    full_tag: Annotated[
+        str | None,
+        Parameter(
+            group=docker_parameters,
+            help="Full tag of the docker image to use. "
+            "If provided, it takes precedence over the version parameter. ",
+        ),
+    ] = None,
     memory: Annotated[
         str | None,
         Parameter(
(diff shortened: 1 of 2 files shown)


ptoupas (approved):
LGTM

modelconverter/__main__.py
```diff
@@ -574,6 +574,17 @@ def launcher(
             "Available options differ based on the target platform. ",
         ),
     ] = None,
+    image: Annotated[
+        str | None,
+        Parameter(
+            ["image", "docker-image"],
+            group=docker_parameters,
+            help="Full name of the docker image to use. "
+            "If the name includes a tag (e.g. ':latest'), "
```
klemen1999, line 583: So is now `:latest` used or is it `:<version>-latest` that needs to be used? Or are both supported? Does it depend on target device (different if  we work with RVC2 which pulls public images vs RVC4 where we have local images)?

ptoupas: The default is still the `:<version>-latest` or `:<version>-dev` as a tag. The only change is that now the user can completely bypass that by setting the `--image/docker-image` flag, which if contains the image name + image tag it will be used as is.
So for example if the `--docker-image luxonis/modelconverter-rvc4:my-custom-tag` is used it will try to find exactly that image with that tag.

klemen1999 (approved):

PR #195, opened 2026-01-09
Added auto-download of SNPE if used locally

klemen1999:
## Purpose

When modelconverter is used locally you can specify tool version. For RVC4 specifically you had to previously download the .zip manually from Qualcomm and put it into the correct place. Since Qualcomm now publicly distributes their SDK we can auto-download it from their catalog if the zip is not present locally.
Note: Version must match exactly for auto-download to work (eg. `2.41.0.251128` works but just `2.41.0` won't work)

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,7 +197,22 @@ Only the version `2022.3.0` of `OpenVino` is supported for `RVC3`. Follow the sa
 
 **RVC4**
 
-Requires `snpe-<version>.zip` archive to be present in `docker/extra_packages`. You can download different SNPE versions from [here](https://softwarecenter.qualcomm.com/catalog/item/Qualcomm_AI_Runtime_Community). After downloading, rename the archive according to the version number. For example, if you download `version 2.32.6`, rename the file to `snpe-2.32.6.zip` and then place it in the `docker/extra_packages` directory.
+Requires `snpe-<version>.zip` archive to be present in `docker/extra_packages`. When building locally via the CLI, the tool will attempt to download the archive automatically if it is missing, as long as the version you pass matches one of the versions available in the Qualcomm catalog.
+You can also download different SNPE versions manually from [here](https://softwarecenter.qualcomm.com/catalog/item/Qualcomm_AI_Runtime_Community). After downloading, rename the archive according to the version number and place it in the `docker/extra_packages` directory.
+
+Example (auto-download on first build):
+
+```bash
+modelconverter convert rvc4 --tool-version 2.32.6.250402 --path <config_or_archive>
+```
+
+Example (archive is already present, no auto-download):
+
+```bash
+# Place the archive ahead of time:
+# docker/extra_packages/snpe-2.40.0.zip
+modelconverter convert rvc4 --tool-version 2.40.0 --path <config_or_archive>
+```
 
 **HAILO**
 
--- a/README.md
+++ b/README.md
@@ -197,21 +197,21 @@ Only the version `2022.3.0` of `OpenVino` is supported for `RVC3`. Follow the sa
 
 **RVC4**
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -197,7 +197,22 @@ Only the version `2022.3.0` of `OpenVino` is supported for `RVC3`. Follow the sa
...
 
-Requires `snpe-<version>.zip` archive to be present in `docker/extra_packages`. You can download different SNPE versions from [here](https://softwarecenter.qualcomm.com/catalog/item/Qualcomm_AI_Runtime_Community). After downloading, rename the archive according to the version number. For example, if you download `version 2.32.6`, rename the file to `snpe-2.32.6.zip` and then place it in the `docker/extra_packages` directory.
+Requires `snpe-<version>.zip` archive to be present in `docker/extra_packages`. When building locally via the CLI, the tool will attempt to download the archive automatically if it is missing, as long as the version you pass matches one of the versions available in the Qualcomm catalog.
+You can also download different SNPE versions manually from [here](https://softwarecenter.qualcomm.com/catalog/item/Qualcomm_AI_Runtime_Community). After downloading, rename the archive according to the version number and place it in the `docker/extra_packages` directory.
+
+Example (auto-download on first build):
+
+```bash
+modelconverter convert rvc4 --tool-version 2.32.6.250402 --path <config_or_archive>
+```
```
ptoupas: Maybe we should add a note here to explicitly state that the whole version `2.32.6.250402` is needed **only** on first time to auto-download and if someone wants to use the generated image afterwards they just need to use the `2.32.6` instead.

modelconverter/utils/docker_utils.py
```diff
@@ -6,15 +6,17 @@
 import tempfile
 from pathlib import Path
 from typing import Literal
+from urllib.error import HTTPError, URLError
+from urllib.request import Request, urlopen
```
ptoupas, line 11: Should we add this package as a new requirement in the `requirements.txt`?

ptoupas (approved):
Overall LGTM, I've just left a couple of comments to be addressed.
One more note is that the size of the `.zip` file that is auto-downloaded seems to be much smaller than the respective file from Qualcomm's software center. It might worth checking into that as well.

README.md
klemen1999: Not really true, you have to specify the full version always since it builds the image like this:
`luxonis/modelconverter-rvc4:2.32.6.250402-latest`
And I would leave it like this honestly.

modelconverter/utils/docker_utils.py
klemen1999: This one is part of the standard libraries

README.md
ptoupas: Its just a bit different from what we've been using so far with the `--tool-version` flag, since we've just using the `2.32.6` part or `2.42.0` etc. But if the docker images are build with the full version then I suppose we should leave it like that.

klemen1999: Yeah the original zip names are with "build" number added. I suppose we could, if user specfies build number ignore it in the local check and use it only for download. This would mean that:
- `2.32.6.250402` is needed the first time because we need to download it
- `2.32.6` works in subsequent runs since the image in previous run would be built with 2.32.6
- `2.32.6.250402` works as well in subsequent runs since we ignore the build number for local check

Thoughts on this?

ptoupas: I think this makes sense. And that way we keep the `--tool-version` usage similar as before but we also allow the use of the new full version numbers if the users decide to use that instead.

klemen1999: Addressed in https://github.com/luxonis/modelconverter/pull/195/commits/33c0d6419388fa8b20a4f530c7330d11865f5ba9, you can try it out now

ptoupas: Tested it and it works as expected.

PR #202, opened 2026-02-12
Automatic docker build

kozlov721:
## Purpose

Makes it possible to fully streamline building a docker image for RVC4 by automatically downloading the SNPE files and modelconverter sources and building the image without the need to clone this repository.

## Specification

- Adds `prepare_build_environment` function that pulls necessary files and creates a build environment for building an RVC4 docker image if it doesn't exist locally and the user is not authorized to pull it from our package repository.

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable


modelconverter/utils/docker_utils.py
```diff
@@ -180,8 +185,46 @@ def docker_build(
...
-    archive_path = Path("docker/extra_packages") / f"snpe-{version}.zip"
+def prepare_build_environemnt(
+    target: Literal["rvc2", "rvc3", "rvc4", "hailo"], version: str
+) -> Path:
+    if target != "rvc4":
+        raise NotImplementedError(
+            "Fully automatic docker build is only implemented for RVC4"
+        )
+
+    build_path = Path(".build", target)
```
kozlov721, line 196: Unsure on how we want to name the build directory and if we would want to keep it hidden or not.

modelconverter/utils/docker_utils.py
```diff
@@ -151,7 +154,9 @@ def docker_build(
 
     tag_version = rvc4_tag_version(version) if target == "rvc4" else version
     if target == "rvc4":
-        ensure_snpe_archive(version)
+        build_dir = prepare_build_environemnt(target, version)
```
ptoupas, line 157: This doesn’t align well with the case where someone has cloned the repo locally and is working with that setup instead of installing from pip.
If the SNPE files are already placed under `docker/extra_packages`, they are ignored, a new `.build` directory is created, and SNPE is automatically downloaded even though they exist.
Maybe we should first check whether `docker/extra_packages` exists (which would indicate a local repo clone workflow) and only create `.build` if it doesn’t (which would indicate a pip-based installation).

modelconverter/utils/docker_utils.py
ptoupas: I believe keeping it hidden inside `.build` is fine.
What I think we should add though is to delete the `modelconverter-{__version__}-beta.zip` file once its extracted and maybe in the extracted `modelconverter-{__version__}-beta` folder delete everything but the `modelconverter-{__version__}-beta/docker` folder.
We don't need to keep all the rest of the files/folders in there since they are not needed or used.

kozlov721: We can add some cleanup but I'd argue that since those are all just tiny source files inside a hidden directory it isn't really worth the extra complexity.

modelconverter/utils/docker_utils.py
kozlov721: Makes sense, fixed

ptoupas (approved):
LGTM

PR #206, opened 2026-02-23
Security: `protobuf` version update

kozlov721:
## Purpose

Fixes a reported `protobuf` [security issue](https://github.com/luxonis/modelconverter/security/dependabot/4)
## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable


klemen1999:
Security alert connected to this is tolerable and was dismissed due to us being unable to fix the problem because Hailo library itself is not compatible witht the fix.

PR #207, opened 2026-02-25
Support for renamed private RVC4 images

kozlov721:
## Purpose

Makes it possible to search in both public and private package repositaries for an RVC4 image.

## Specification

- Image pull is not performed only on the first candidate image, but on all the candidates untill the first one succeeds. 
- Updated CI to push certain RVC4 packages to a package with `-private` suffix 

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable


modelconverter/utils/docker_utils.py
```diff
@@ -359,9 +360,13 @@ def get_docker_image(
         image = f"{image_repo}:{tag}"
 
     candidate_images = [image]
-    # add full version if specified RVC4 tag is with build number included (e.g. version=2.32.6.250402 instead of version=2.32.6)
-    if target == "rvc4" and tag_version != version and image_tag is None:
-        candidate_images.append(f"{image_repo}:{version}-{bare_tag}")
+    if target == "rvc4":
+        candidate_images.append(f"{image_repo}-private:{tag}")
```
klemen1999: What is the difference between `tag` and `bare_tag`? Why do we in "public" case use `{version}-{bare_tag}` and here just `{tag}`

kozlov721: `tag` is equivalent to `{version}-{bare_tag}`. `bare_tag` is just one of `"latest"` or `"dev"`.

klemen1999: Any reason then why we don't use `tag` for the public one?

kozlov721: `version` is not the same as `tag_version`. There is a distuinction in the build number I believe, it was added in #195.

klemen1999: Ohh I see now, I had to expand the code a bit further to go through the whole logic. Makes sense now, should be good like this

klemen1999 (commented):
LGTM

klemen1999 (approved):

PR #212, opened 2026-03-14
RVC4: Support for JSON quantization overrides

kozlov721:
## Purpose

Allows users to specify custom quantization overrides using `rvc4.encodings` field.

## Specification

- Added `rvc4.encodings` to the configuration specs.
  - Can be either a dictionary or a path to a JSON file

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/modelconverter/packages/rvc4/exporter.py
+++ b/modelconverter/packages/rvc4/exporter.py
@@ -40,7 +40,7 @@ class RVC4Exporter(Exporter):
         super().__init__(config=config, output_dir=output_dir)
 
         rvc4_cfg = config.rvc4
-        self.custom_quantization_overrides = rvc4_cfg.quantization_overrides
+        self.encodings = rvc4_cfg.encodings
         self.snpe_onnx_to_dlc = rvc4_cfg.snpe_onnx_to_dlc_args
         self.snpe_dlc_quant = rvc4_cfg.snpe_dlc_quant_args
         self.snpe_dlc_graph_prepare = rvc4_cfg.snpe_dlc_graph_prepare_args
@@ -187,7 +187,7 @@ class RVC4Exporter(Exporter):
             args.append("--override_params")
         elif self.quantization_mode == QuantizationMode.INT8_16_MIX:
             self._add_args(args, ["--act_bitwidth", "16"])
-        elif self.custom_quantization_overrides is not None:
+        elif self.encodings is not None:
             args.append("--override_params")
 
         start_time = time.time()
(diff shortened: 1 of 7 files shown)


modelconverter/utils/config.py
```diff
@@ -265,6 +261,28 @@ class RVC3Config(BlobBaseConfig):
     pot_target_device: PotDevice = PotDevice.VPU
 
 
+class QuantizationOverridesItem(BaseModelExtraForbid):
+    bitwidth: Annotated[int, Field(ge=4, le=32)] | None = None
```
ptoupas: Since only specific bitwidths are supported, it might be better to use a Literal instead of a ranged integer.
`bitwidth: Literal[4, 8, 16, 32] | None = None`

modelconverter/utils/config.py
```diff
@@ -265,6 +261,28 @@ class RVC3Config(BlobBaseConfig):
...
+    @staticmethod
+    def serialize_is_symmetric(value: bool | None) -> str | None:
+        if value is None:
+            return None
+        return str(value)
+
+
+class QuantizationOverrides(BaseModelExtraForbid):
+    activation_encodings: dict[str, list[QuantizationOverridesItem]]
+    parameter_encodings: dict[str, list[QuantizationOverridesItem]]
```
ptoupas: This field needs to be renamed into `param_encodings` as expected by the SNPE, and as described [here](https://docs.qualcomm.com/doc/80-63442-10/topic/quantization.html#quantization-overrides)

modelconverter/packages/rvc4/exporter.py
ptoupas: We'd have to update this to account for the case the `rvc4.quantization_overrides` flag is provided so that the encodings are actually applied to the SNPE command, otherwise they can never be applied.
```suggestion
        elif (
            self.quantization_mode
            in [
                QuantizationMode.INT8_16_MIX,
                QuantizationMode.INT8_16_MIX_ACC,
            ]
            or self.custom_quantization_overrides is not None
        ):
```

ptoupas, line 191: We'd have to add an extra statement here to account for the case the rvc4.quantization_overrides flag is provided so that the `--override_params` is enabled for the `snpe-dlc-quant` so that the encodings can actually be applied.
```suggestion
        elif self.custom_quantization_overrides is not None:
            args.append("--override_params")
```

ptoupas (changes requested):
Left a few comments and need to be addressed before merging this.

modelconverter/packages/rvc4/exporter.py
```diff
@@ -273,19 +275,27 @@ class Entry(NamedTuple):
...
+                )
+            for name in (
+                list(self.inputs.keys())
+                + list(self.outputs.keys())
+                + self.extra_quant_tensors
+            ):
+                encodings_dict["activation_encodings"][name] = [
+                    {"bitwidth": 8, "dtype": "int"}
+                ]
         encodings_path = self.intermediate_outputs_dir / "io_encodings.json"
```
ptoupas: Lets update that to `encodings.json` as well.
```suggestion
        encodings_path = self.intermediate_outputs_dir / "encodings.json"
```

ptoupas (approved):
LGTM

PR #216, opened 2026-04-08
Bumped ONNX to 1.21

klemen1999:
## Purpose

- Bumped ONNX to `1.21.0` in `requirements.txt`.
- Added onnx_compatibility.py to isolate `ONNX 1.21` compatibility helpers.
- Patched missing ONNX helper conversions before importing `GraphSurgeon` so `onnx_graphsurgeon` continues to work with `ONNX 1.21`.
- Centralized ONNX model saving, including external-data handling, to avoid repeated save logic and sidecar overwrite issues.
- Made ONNX simplification fallback gracefully when `onnxsim` fails instead of aborting export.
- Tightened ONNX shape parsing so symbolic or missing dimensions are handled explicitly instead of silently reading as 0.
- Extended ONNX dtype support only for the additional types needed by current code paths: `BFLOAT16`, `INT4`, and `UINT4`.
- Added focused compatibility regression coverage in test_onnx_compatibility.py.

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/modelconverter/utils/onnx_compatibility.py
+++ b/modelconverter/utils/onnx_compatibility.py
@@ -3,7 +3,6 @@ from pathlib import Path
 import ml_dtypes
 import numpy as np
 import onnx
-from onnx.external_data_helper import convert_model_to_external_data
 
 
 def ensure_onnx_helper_compatibility() -> None:
@@ -51,13 +50,6 @@ def save_onnx_model(
         )
         if external_data_path.exists():
             external_data_path.unlink()
-        convert_model_to_external_data(
-            model,
-            all_tensors_to_one_file=True,
-            location=external_data_path.name,
-            size_threshold=0,
-            convert_attribute=False,
-        )
         onnx.save(
             model,
             str(output_path),
--- a/modelconverter/utils/onnx_compatibility.py
+++ b/modelconverter/utils/onnx_compatibility.py
@@ -1,0 +1,72 @@
+from pathlib import Path
+
+import ml_dtypes
+import numpy as np
+import onnx
+from onnx.external_data_helper import convert_model_to_external_data
+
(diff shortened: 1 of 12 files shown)


klemen1999:
This PR is mostly introduced because of vulnerabilities in ONNX<=1.20.0. We still have to use lower ONNX versions in some parts (RVC4 and Hailo) but we don't have control over that.

kozlov721 (approved):
LGTM

modelconverter/utils/onnx_compatibility.py
```diff
@@ -0,0 +1,72 @@
...
+        )
+        if external_data_path.exists():
+            external_data_path.unlink()
+        convert_model_to_external_data(
+            model,
+            all_tensors_to_one_file=True,
+            location=external_data_path.name,
+            size_threshold=0,
+            convert_attribute=False,
+        )
```
ptoupas: This function call seems redundant and might even cause issues in certain cases.
The following command, i.e. `onnx.save` with the  `save_as_external_data=True` should internally call the exact same function, so we should probably remove it.

ptoupas (approved):
Apart from my [comment](https://github.com/luxonis/modelconverter/pull/216/changes#r3077990381) on the redundant call of the `convert_model_to_external_data` it looks good to me.

modelconverter/utils/onnx_compatibility.py
klemen1999: Good catch, addressed in https://github.com/luxonis/modelconverter/pull/216/commits/d95540c3a7a76a87f21e3b6ff7bcbba051f74bbe

PR #219, opened 2026-04-20
Generalized `AdbHandler` to work over SSH

kozlov721:
## Purpose

Makes it possible to control devices that are only accessible over SSH.

## Specification

- Created abstract `DeviceHandler` and added new `SSHHandler` class
- New helper `create_handler` function

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable


modelconverter/packages/base_benchmark.py
```diff
@@ -1,23 +1,35 @@
...
+    system_power_peak: float | None = None
+    processor_power_mean: float | None = None
+    processor_power_median: float | None = None
+    processor_power_peak: float | None = None
+    dsp_mean: float | None = None
+    dsp_median: float | None = None
+    dsp_peak: float | None = None
+    memory_mean: float | None = None
+    memory_median: float | None = None
+    memory_peak: float | None = None
```
ptoupas: Image

I think adding all of these extra results makes the output in the terminal way too crowded and hard to focus on the metrics that matter the most. Further more its getting packed and its not properly showing all the information (e.g. its cutting of the words and the descriptions of metrics).
My suggestion is to keep that terminal output leaner, keeping only the `_mean` values and maybe create a csv or something where we report everything else (the expanded metrics).
Thoughts @kozlov721, @klemen1999?

modelconverter/packages/rvc4/benchmark.py
```diff
@@ -301,21 +298,35 @@ def benchmark(self, configuration: Configuration) -> BenchmarkResult:
...
+                    )
+                    result.processor_power_peak = stats.power_processor_w.peak
+                if stats.dsp_util_percent is not None:
+                    result.dsp_mean = stats.dsp_util_percent.mean
+                    result.dsp_median = stats.dsp_util_percent.median
+                    result.dsp_peak = stats.dsp_util_percent.peak
+                if stats.ram_used_mb is not None:
+                    result.memory_mean = stats.ram_used_mb.mean
+                    result.memory_median = stats.ram_used_mb.median
+                    result.memory_peak = stats.ram_used_mb.peak
```
ptoupas:
```suggestion
            if self.monitor:
                stats = self.monitor.get_stats()
                stat_mapping = {
                    "power_system_w":    "system_power",
                    "power_processor_w": "processor_power",
                    "dsp_util_percent":  "dsp",
                    "ram_used_mb":       "memory",
                }
                for stat_attr, result_prefix in stat_mapping.items():
                    stat = getattr(stats, stat_attr)
                    if stat is None:
                        continue
                    for metric in ("mean", "median", "peak"):
                        setattr(result, f"{result_prefix}_{metric}", getattr(stat, metric))
```
I think we could make this part a bit more compact and with less redundant and repetitive code.

modelconverter/packages/base_benchmark.py
klemen1999: Agree, let's only include the mean but maybe have a --verbose flag for full output in CLI and potentially --output for file save

ptoupas (approved):
LGTM

PR #220, opened 2026-04-23
Upgraded `DeviceMonitor`

kozlov721:
## Purpose

Allows to monitor more metrics

## Specification

- Added RAM and CPU monitoring
- Temperature monitoring 
- DSP frequencies instead of just the overall utilization
- Merged all the monitors into a single `DeviceMonitor` 

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable


modelconverter/packages/base_benchmark.py
```diff
@@ -1,23 +1,38 @@
...
 from loguru import logger
 
 from modelconverter.utils import is_hubai_available, resolve_path
 
 
-class BenchmarkResult(NamedTuple):
+@dataclass
+class BenchmarkResult:
     """Benchmark result, tuple (FPS, latency in ms, system and processor
     power in W, dsp utilization)"""
```
ptoupas: Lets update that docstring respectively to the Class changes.

modelconverter/packages/rvc4/benchmark.py
```diff
@@ -504,26 +518,31 @@ def _extra_header(
         results: list[tuple[Configuration, BenchmarkResult]],
     ) -> list[str]:
         heads = []
-        if self.power_monitor:
+        if self.monitor:
             heads.append("power_sys (W)")
             heads.append("power_core (W)")
-        if self.dsp_monitor:
             heads.append("dsp (%)")
+            heads.append("memory (MB)")
```
ptoupas: We should probably change this to `MiB` since we return the memory in MiB not in MB (divide by 1024) in the [device_monitors.py](https://github.com/luxonis/modelconverter/blob/49e63cacdb75e45f3d2895133d82f65746461c5c/modelconverter/utils/device_monitors.py#L288).
Lets try to be consistent on this annotation.

modelconverter/utils/device_monitors.py
```diff
@@ -88,46 +87,124 @@
...
+            float | None,
+            float | None,
+        ]
+        | None
+    ):
+        system, proc = self.read_power()
+        dsp = self.read_dsp()
+        ram = self.read_ram()
+        cpu = self.read_cpu()
+        return system, proc, dsp, ram, cpu
```
ptoupas: Return type annotation is wrong. Either drop the | None, or return None if all reads failed.

modelconverter/utils/device_monitors.py
```diff
@@ -138,272 +215,200 @@ def start(self) -> None:
...
+        self.idle_power_system = stats["power_system"] or 0.0
+        self.idle_power_processor = stats["power_processor"] or 0.0
+        self.idle_ram_used = stats["ram_used"] or 0.0
+        self.idle_dsp_utilization = stats["dsp"] or 0.0
+        self.idle_cpu_utilization = stats["cpu"] or 0.0
+
+        self.idle_power_system *= 1.1
+        self.idle_power_processor *= 1.1
+        self.idle_dsp_utilization *= 1.1
+        self.idle_cpu_utilization *= 1.1
```
ptoupas: Is there a specific reason we don't give the `self.idle_ram_used` the same `1.1×` margin as the rest of the idle values?

modelconverter/utils/device_monitors.py
```diff
@@ -138,272 +215,200 @@ def start(self) -> None:
...
             return None
+        code, dsp, error = self.device_handler.shell(
+            "bash /data/modelconverter/oak_dsp_util.sh", check=False
+        )
+        if "Maximum frequency is zero" in error:
+            dsp = 0
+        elif code != 0:
+            logger.warning("Failed to read DSP value.")
+            return None
+        return float(dsp)
```
ptoupas: This casting to float might fail if the dsp value from above (line 358) returns garbage. Better to validate before the cast.

ptoupas (commented):
A couple extra comments on top of the on-code comments I provided:

1. Setting just the `--device-id ...` flag on a benchmark run (e.g. `modelconverter benchmark rvc4 --model-path luxonis/yolov8-nano-pose-estimation:coco-512x288 --power-benchmark --dsp-benchmark --device-id da79b2bf`) throws the following warning: `WARNING  No device ID specified, using the first connected device: bb81a24e.` which shouldn't be an issue given that we are providing the exact device-id. On the other hand if just the `--device-ip` is provided the `--device-id` is inferred and found properly
2. It seems like if any of the `--power-benchmark`, `--dsp-benchmark` flags are provided in CLI the other one is also enabled by default. Even if I explicitly set the other one to `--no-...` the respective metric is tracked while it shouldn't. For example the following flags track both power and dsp `--power-benchmark --no-dsp-benchmark`. Same if I just set one like `--power-benchmark` or `--dsp-benchmark` both are tracked while it shouldn't.

kozlov721:
> Setting just the --device-id ... flag on a benchmark run 

This is because the `device-id` here does not refer to the adb ID (which is in hex) but the regular ID (decimal). I added the support for the adb ID format too so both are accepted now

ptoupas (approved):
Reviewed and re-tested the new updates. Everything seems to be addressed now.
LGTM 🚀

klemen1999 (approved):

PR #222, opened 2026-04-23
Support multi-input model benchmarking

klemen1999:
## Purpose

- Added support for benchmarking multi input models by getting the input types from the .dlc directly

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

Tested multi input model with:
`modelconverter benchmark rvc4 --model-path luxonis/whisper-tiny-en:decoder --device-ip ...`


modelconverter/packages/rvc4/benchmark.py
```diff
@@ -175,81 +187,39 @@ def _get_data_type_from_dlc(self) -> dai.TensorInfo.DataType:
...
+        input_size: list[int], data_type: dai.TensorInfo.DataType
+    ) -> np.ndarray:
+        if data_type == dai.TensorInfo.DataType.FP32:
+            return np.random.rand(*input_size).astype(np.float32)
+        if data_type == dai.TensorInfo.DataType.FP16:
+            return np.random.rand(*input_size).astype(np.float16)
+        if data_type == dai.TensorInfo.DataType.FP64:
+            return np.random.rand(*input_size).astype(np.float64)
+        if data_type == dai.TensorInfo.DataType.I8:
+            return np.random.randint(-128, 127, size=input_size, dtype=np.int8)
```
ptoupas: Minor one but this should be -128, 128 (to allow the 127 value to actually be  used)

modelconverter/packages/rvc4/benchmark.py
```diff
@@ -175,81 +187,39 @@ def _get_data_type_from_dlc(self) -> dai.TensorInfo.DataType:
...
+        if data_type == dai.TensorInfo.DataType.INT:
+            # INT inputs are often token/index tensors; zero keeps synthetic
+            # benchmark inputs in a safe range.
+            return np.zeros(input_size, dtype=np.int32)
+        if data_type == dai.TensorInfo.DataType.U8F:
+            return np.random.randint(0, 255, size=input_size, dtype=np.uint8)
```
ptoupas: Same as above, this should be 0, 256 (to allow the 255 value to actually be  used)

modelconverter/packages/rvc4/benchmark.py
```diff
@@ -45,6 +44,16 @@
     "cpu": "use_cpu",
 }
 
+DLC_TO_DAI_DATA_TYPE: Final[dict[str, dai.TensorInfo.DataType]] = {
+    "Float_32": dai.TensorInfo.DataType.FP32,
+    "Float_16": dai.TensorInfo.DataType.FP16,
+    "Float_64": dai.TensorInfo.DataType.FP64,
+    "Int_8": dai.TensorInfo.DataType.I8,
+    "Int_32": dai.TensorInfo.DataType.INT,
+    "uFxp_16": dai.TensorInfo.DataType.FP16,
```
ptoupas, line 56: Since this is a fixed point 16 and not a float16 should't we map that to something else for DAI? I know that INT16 is currently not supported in DAI, so maybe we should add a comment here to update it when DAI supports the INT16.
I am not sure if mapping that to FP16 is the right thing to do.

ptoupas (commented):
Apart from the in-code comments I've put down, there is one more below:
I think we should update the SNPE-based benchmark to also use the new way of getting the model inputs instead of keep using the old one (`_prepare_raw_inputs`).

modelconverter/packages/rvc4/benchmark.py
klemen1999: Good catch, addressed in https://github.com/luxonis/modelconverter/pull/222/commits/4b8bbe34bcb47f6c50acd59659e4c4d1a18c4b14

modelconverter/packages/rvc4/benchmark.py
klemen1999: Addressed in https://github.com/luxonis/modelconverter/pull/222/commits/4b8bbe34bcb47f6c50acd59659e4c4d1a18c4b14

modelconverter/packages/rvc4/benchmark.py
klemen1999: Yeah agree, FP16 is not the best. I added a comment to this line to fix in the future but for now we'll leave it at this. We did treat `uFxp_16` as `np.float16` in the SNPE conversion as well so hopefully this doesn't cause much issues. And DAI support for INT16 is in development, should come out with next release anyway.

klemen1999:
> I think we should update the SNPE-based benchmark

Good point, I added this now in as well

ptoupas (approved):
LGTM

PR #230, opened 2026-05-15
[Benchmark] Guess correct input dtypes

kozlov721:
## Purpose

Improved ability to guess the correct input types in case the information contained in the NN Archive is incorrect.

## Specification

- Reverted some changes made in #222 
  - The `precision` field from the HubAI `GET /modelInstance` response is usually the most trustworthy
  - If the dtype is too ambiguous, the dtypes are read from the DLC itself 

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

## AI Usage

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable input-spec derivation: primary extraction is attempted first with a safe fallback when it fails.
  * Improved error handling for external command execution to surface failures consistently.

* **Refactor**
  * Enhanced input data-type inference by incorporating model precision and optional metadata lookups for more accurate benchmarks.

* **Chores**
  * Cleaned up a removed, no-longer-exported exception to streamline error utilities.


modelconverter/packages/rvc4/benchmark.py
```diff
@@ -457,18 +464,114 @@ def _extra_row_cells(
...
+                    "model_precision_type", "INT8"
+                )
+                break
+
+        if model_precision_type == "FP16":
+            return DataType.FLOAT16
+        if model_precision_type == "FP32":
+            return DataType.FLOAT32
+
+        return DataType.INT8
```
klemen1999: Let's explicitly check for INT8 and if it doesn't match any of the cases explicitly error out. This is because shortly Hub will add mixed precision types and our backend will assume INT8 always with this code

modelconverter/packages/rvc4/benchmark.py
```diff
@@ -457,18 +464,114 @@ def _extra_row_cells(
...
-    def _get_archive_input_specs(archive: dai.NNArchive) -> list[InputSpec]:
+    def _get_archive_input_specs(
+        self, archive: dai.NNArchive
+    ) -> list[InputSpec]:
+        def guess_dtype(
+            dtype: DataType,
+            archive_precision: DataType | None,
+            hubai_precision: DataType | None = None,
+        ) -> DataType:
+            if dtype is DataType.INT32:
```
klemen1999: Where would this case apply?

modelconverter/packages/rvc4/benchmark.py
```diff
@@ -457,18 +464,114 @@ def _extra_row_cells(
...
+                return DataType.INT32
+            if hubai_precision is not None:
```
klemen1999, line 496: Let's have some log when this happens so it is clearn from the CLI that precision was taken from HubAI "metadata". And similarly, let's log if precision was taken from the NNArchive metadata

modelconverter/packages/rvc4/benchmark.py
```diff
@@ -349,11 +350,17 @@ def _benchmark_dai(
 
         input_specs: list[InputSpec] = []
         if isinstance(model_path, str) or str(model_path).endswith(".tar.xz"):
-            model_archive = dai.NNArchive(modelPath)  # type: ignore[arg-type]
-            if shutil.which("snpe-dlc-info") is not None:
-                input_specs = self._get_dlc_input_specs(modelPath)
-            else:
+            model_archive = dai.NNArchive(modelPath)
```
klemen1999, line 371: I would still have this switched around so that we first try to read from DLC directly using SNPE tools and if that fails have the archive spec as a backup. Reason being that reading from DLC directly is 100% guranteed, other paths are not

modelconverter/packages/rvc4/benchmark.py
kozlov721: If the input is an index. I added a comment.

klemen1999 (approved):

PR #237, opened 2026-06-05
Fix failing SNPE benchmark

kozlov721:
## Purpose

Fixed various issues with RVC4 benchmarking on SNPE backend.
 
## Specification

- Fixed using an empty input list causing Segmentation fault
- Added more logging for the SNPE benchmark
- Sped up the preparation of raw data
  - All files are pushed at once 
  - Only 100 files are created, the rest are hard links to prevent the device from running out of memory (the raw data can be pretty large)
  - Decreased the default number of images to 500 to lower the chance of OOM on the default settings 
- Postponed the start of device monitoring so it does not run during data preparation 
- Results table is now rendered vertically instead of horizontally
- Possible OOM memory errors are now detected and the user is advised to decrease the number of images
- Added option to pass `silent` to `DeviceHandler.shell` for easier debugging
- Removed duplicate docs in device handlers (inherited through `@override`)
- Removed latency reporting from `depthai` because reported number is incorrect due to a lack of support
- `snpe-dlc-info` can be either on the device or on the host
- Expanded README

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

## AI Usage

## Summary by CodeRabbit

* **Improvements**
  * Faster, bulk staging of benchmark inputs with progress feedback and a cap on generated inputs; safer linking for extras.
  * Prefer local DLC inspection when available, with automatic fallback to remote extraction.
  * Benchmark results table transposed for clearer metric-by-metric view; adjusted image-count defaults.

* **Bug Fixes**
  * Improved benchmark lifecycle and logging; explicit handling for killed runs; dai-benchmark reports latency as "N/A" when appropriate.

* **Documentation**
  * Clarified and reorganized README guidance for configuration, running, and RVC4 usage.

* **Style**
  * Configurable command verbosity to reduce noisy device output.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ There are two main ways to execute configure the conversion process:
 
 1. **NN Archive**:
    Alternatively, you can use an [NN Archive](https://docs.luxonis.com/software-v3/ai-inference/nn-archive/#NN%20Archive) as input. An NN Archive includes a model in one of the supported formats—ONNX (.onnx), OpenVINO IR (.xml and .bin), or TensorFlow Lite (.tflite)—alongside a `config.json` file. The config.json file follows a specific configuration format as described under the `Configuration` section.
-1. **YAML Configuration File (Legacy)**:
+2. **YAML Configuration File (Legacy)**:
    An alternative way to configure the conversion is through a YAML configuration file. For reference, you can check [defaults.yaml](shared_with_container/configs/defaults.yaml) and other examples located in the [shared_with_container/configs](shared_with_container/configs) directory.
 
 **Modifying Settings with Command-Line Arguments**:
@@ -359,11 +359,11 @@ You can run the built image either manually using the `docker run` command or us
    export AWS_S3_ENDPOINT_URL=<your_aws_s3_endpoint_url>
    ```
 
-1. If `shared_with_container` directory doesn't exist on your host, create it.
+2. If `shared_with_container` directory doesn't exist on your host, create it.
 
-1. Without remote files, place the model, config, and calibration data in the respective directories (refer [Sharing Files](#sharing-files)).
+3. Without remote files, place the model, config, and calibration data in the respective directories (refer [Sharing Files](#sharing-files)).
 
-1. Execute the conversion:
+4. Execute the conversion:
 
 - If using the `modelconverter` CLI:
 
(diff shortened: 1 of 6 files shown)


modelconverter/packages/rvc4/benchmark.py
```diff
@@ -158,12 +159,13 @@ def _prepare_raw_inputs(
...
             f.write(input_list.encode())
-            self.handler.push(
-                f.name,
-                f"/data/modelconverter/{self.model_name}/input_list.txt",
-            )
+
+        self.handler.push(
+            f.name,
+            f"/data/modelconverter/{self.model_name}/input_list.txt",
+        )
```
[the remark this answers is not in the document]
klemen1999: @kozlov721 let's handle this

klemen1999 (approved):

README.md
```diff
@@ -611,38 +611,34 @@ modelconverter benchmark rvc4 --model-path <path_to_model.tar.xz>
...
-> To access models from different teams in Luxonis HubAI, remember to update the `HUBAI_API_KEY` environment variable accordingly.
+> To access models from different teams in Luxonis HubAI, remember to update the `HUBAI_API_KEY` environment variable accordingly when using a model slug as an input.
 
 > [!NOTE]
-> **Benchmark Duration Control (RVC2/RVC4)**: Two flags can affect the duration of benchmarking:
->
-> - `--benchmark-time`: Duration in seconds for time-based benchmarking (default: 20)
-> - `--repetitions`: Number of iterations to perform (default: 10)
->
-> By default, the benchmarking uses `--benchmark-time` (20 seconds) which takes precedence over `--repetitions`. To use `--repetitions` instead, you must explicitly set `--benchmark-time` to a negative value (e.g., `--benchmark-time -1`).
```
klemen1999, line 629: I'd keep this information in somewhere in the README, wouldn't delete it

modelconverter/__main__.py
```diff
@@ -330,7 +330,7 @@ def benchmark(
         Parameter(group="RVC4"),
     ] = "balanced",
     runtime: Annotated[Literal["dsp", "cpu"], Parameter(group="RVC4")] = "dsp",
-    num_images: Annotated[int, Parameter(group="RVC4")] = 1000,
+    num_images: Annotated[int, Parameter(group="RVC4")] = 500,
```
klemen1999, line 333: [Here](https://github.com/luxonis/modelconverter/blob/8185b4f252aca895273e22ea6680293cc2888a90/modelconverter/packages/rvc4/benchmark.py#L70) we still have default to 1000 images.

modelconverter/packages/rvc4/benchmark.py
```diff
@@ -288,23 +350,37 @@ def _benchmark_snpe(
...
             f"--container /data/modelconverter/{self.model_name}/model.dlc "
             f"--input_list /data/modelconverter/{self.model_name}/input_list.txt "
             f"--output_dir /data/modelconverter/{self.model_name}/outputs "
             f"--perf_profile {profile} "
             "--cpu_fallback true "
-            f"--{runtime}"
+            f"--{runtime}",
+            check=False,
         )
+        if retcode == 137:
```
klemen1999: Is this the only retcode that we should care about? Because now that snpe-parallel-run has check=False it seems like there could be another error but would be silenced and not user wouldn't know that anything bad happened

modelconverter/utils/device_handlers.py
klemen1999, line 1: We have some bugs in the SCP handler:
- In the `push()` command the scp command that gets built is `-r` has attached to the end. It should be `scp -r ...` fo rit to actually work recursively
- same thing for the `pull()` command under the SCP handler (-r should be moved to the front of the command)

klemen1999 (commented):
Couple of comments

modelconverter/utils/device_handlers.py
kozlov721: Afaik that shouldn't matter but I'll double check or just move it to the front anyways

modelconverter/packages/rvc4/benchmark.py
kozlov721: Oh right I missed this, thanks

modelconverter/utils/device_handlers.py
klemen1999: I think it didn't matter before because we didn't use it as a recursive tool. Though might be a bit implementation dependant if -r at the end works but overall I think for safety we should move it to the front

klemen1999 (approved):

PR #241, opened 2026-06-08
Support for `onnxslim`

kozlov721:
## Purpose

`onnxslim` can slightly increase the performance of the model.

## Specification

- Removed option `disable_onnx_simplification`
- Added `onnx_simplification` which can take either `"onnxsim"` (default), `"onnxslim"`, or `False`
  - Setting to `False` equivalent to `disable_onnx_simplification=True` 

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

## AI Usage

## Summary by CodeRabbit

* **New Features**
  * ONNX model simplification now supports selectable backends (default "onnxsim"), an experimental "onnxslim", or disabling simplification.

* **Bug Fixes**
  * Download fallback: when directory checks are forbidden, single-file downloads proceed instead of failing.

* **Chores**
  * Replaced deprecated boolean setting with a new multi-value configuration; legacy setting auto-migrates with a deprecation warning and optional backend packages added.

* **Tests**
  * Updated defaults and tests to cover the new simplification option and the download-fallback behavior.

### DIFF
--- a/modelconverter/packages/rvc4/exporter.py
+++ b/modelconverter/packages/rvc4/exporter.py
@@ -34,9 +34,6 @@ class RVC4Exporter(Exporter):
     target: Target = Target.RVC4
 
     def __init__(self, config: SingleStageConfig, output_dir: Path):
-        # NOTE: Handled inside `snpe-onnx-to-dlc` command.
-        config.disable_onnx_simplification = True
-
         super().__init__(config=config, output_dir=output_dir)
 
         rvc4_cfg = config.rvc4
--- a/modelconverter/utils/config.py
+++ b/modelconverter/utils/config.py
@@ -362,7 +362,7 @@ class SingleStageConfig(BaseModelExtraForbid):
     def validate_onnx_simplification(
         cls, data: dict[str, Any]
     ) -> dict[str, Any]:
-        if data.get("disable_onnx_simplification", False):
+        if data.pop("disable_onnx_simplification", False):
             warnings.warn(
                 "`disable_onnx_simplification` is deprecated. Please use "
                 "`onnx_simplification` set to `False` instead.",
--- a/modelconverter/utils/config.py
+++ b/modelconverter/utils/config.py
@@ -363,6 +363,12 @@ class SingleStageConfig(BaseModelExtraForbid):
         cls, data: dict[str, Any]
     ) -> dict[str, Any]:
         if data.pop("disable_onnx_simplification", False):
+            if "onnx_simplification" in data:
+                raise ValueError(
+                    "Cannot specify both `disable_onnx_simplification` "
+                    "and `onnx_simplification`."
+                )
+
             warnings.warn(
                 "`disable_onnx_simplification` is deprecated. Please use "
                 "`onnx_simplification` set to `False` instead.",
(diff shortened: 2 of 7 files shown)


modelconverter/utils/config.py
```diff
@@ -345,7 +346,7 @@ class SingleStageConfig(BaseModelExtraForbid):
     outputs: Annotated[list[OutputConfig], Field(min_length=1)] = []
 
     keep_intermediate_outputs: bool = True
-    disable_onnx_simplification: bool = False
```
klemen1999, line 370: Hub also exposes this flag so this is a breaking change for them as well - let's note this down in the PR title that we are removing a parameter.

modelconverter/packages/rvc4/exporter.py
```diff
@@ -34,9 +34,6 @@ class RVC4Exporter(Exporter):
     target: Target = Target.RVC4
 
     def __init__(self, config: SingleStageConfig, output_dir: Path):
-        # NOTE: Handled inside `snpe-onnx-to-dlc` command.
```
klemen1999, line 37: Do you know what was this about?

klemen1999 (approved):
Generally LGTM

modelconverter/packages/rvc4/exporter.py
kozlov721: `snpe-onnx-to-dlc` has a built in support for `onnxsim` simplification so we used to just let it handle it instead of simplifying ourselves. But now when we're exposing a second backend for the simplification this can no longer be done.

modelconverter/utils/config.py
kozlov721: The old option will still be accepted for a while, it will just throw a deprecation warning.

kozlov721:
Perhaps one open question: should we make `onnxslim` the default already? I was thinking of keeping it opt-in for maybe one more release but we can just as well switch to it right away since the next release will be breaking anyways (and this should not even be a breaking change).

modelconverter/packages/rvc4/exporter.py
klemen1999: So I assume now we'll be simplifying twice: Once with our selected pacakge and once by SNPE automatically inside `snpe-onnx-to-dlc`?

klemen1999:
> Perhaps one open question: should we make onnxslim the default already?

Do we have enough data to say that onnxslim is default better than `onnxsim`?

kozlov721:
> Do we have enough data to say that onnxslim is default better than onnxsim?

I'm not 100 % sure about that, we did some benchmarking and it's supposed to be ~0.5% faster but I'll run some more tests and keep it opt-in for now.

modelconverter/packages/rvc4/exporter.py
kozlov721: Yes. Well I suppose the SNPE simplifications wouldn't do anything extra but I can add a flag to SNPE to explicitly disable them.

kozlov721: Or we can add `"native"` as a third option for `onnx_simplification` which would only be supported for RVC4 and if selected only the SNPE simplifications are used (I believe they are just calling `onnxsim` anyways).

klemen1999: I think it's fine to leave as is and run it twice for RVC4. No need to overcomplicate at this point

PR #246, opened 2026-06-12
ONNX modifications with no op exit early

klemen1999:
## Purpose

- If ONNX model doesn't require any modifications then we just return the original model instead of loading the whole graph into memory. This becomes especially noticeable with bigger models
- If there already exists `<model name>.info.csv` file from `snpe-dlc-info` and is not older than the dlc file itself then we use it. This way we save some time instead of always re-generating it

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

## AI Usage

## Summary by CodeRabbit

* **Refactor**
  * Improved ONNX input normalization behavior by skipping model rewriting when no changes are needed.
  * Added clear configuration logic to determine when ONNX input modifications are required.
  * Optimized DLC metadata handling by reusing existing metadata and updating only when the source model is newer.

* **Tests**
  * Added coverage to confirm the no-op ONNX path is returned and no modified output is produced when normalization changes are not required.


dtronmans (approved):
LGTM, thanks. Just has a few merge conflicts though

PR #248, opened 2026-06-13
Removes private handling of RVC4 images

klemen1999:
## Purpose

Removes private RVC4 image package

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

## AI Usage

## Chores
- Simplified container image publishing to use unified, non-private image naming across registries, removing private tag variants.
- Reorganized workflow so the image build happens before registry publishes, with consistent `latest` and commit-SHA tags computed from the same inputs.
- Improved RVC4 Docker candidate selection to refine when “full version” fallbacks are appended.

## Bug Fixes
- Fixed workflow behavior so build files are downloaded consistently, including for the hailo package.


klemen1999:
When looking at the `publish.yaml` I also randomly saw [this hailo specific check](https://github.com/luxonis/modelconverter/blob/cb24581656f7053ca50653160dcc211f8848793d/.github/workflows/publish.yaml#L82). It looks like we skip this download for hailo but the step has hailo specific code. Is it intentional that we skip it anyway? @kozlov721

kozlov721 (approved):

PR #262, opened 2026-07-25
Hide docker and make conversion rootles

kozlov721:
## Purpose

Makes it much easier to use `modelconverter` without managing shared directory with containers and dealing with root being the owner of converted files.

## Specification

- Movement of files between the user and the container is now managed automatically in the background. 
  - Removed `shared_with_container/`
  - User can directly call `modelconverter convert rvc4 --path ~/Downloads/resnet18.onnx.tar.xz`
  - Results will be saved under `output/` in CWD
  - Converted results are owned by the user who invoked the conversion
- Files internally stored in a global cache
  - Lives in `~/.cache/modelconverter` (respects `XDG_CACHE_HOME`), keyed by content hash so an unchanged input is only staged once
  - Held to a size budget via `MODELCONVERTER_CACHE_MAX_SIZE` (default `50GiB`, `0` disables it)
    - Staging ends by evicting least recently used entries until the cache fits again
    - Entries claimed by a running conversion are never evicted, so a parallel run keeps what it staged
    - Superseded copies of a re-staged input are dropped, so re-converting a model being worked on no longer leaves every earlier version behind
  - Added `modelconverter cache` commands to inspect and purge it
    - `cache info` reports the location, per-directory usage and the active budget
    - `cache clean` refuses to run while another conversion is still using the cache

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

## AI Usage

Submitted code was reviewed by a human: YES

The author is taking the responsibility for the contribution: YES

## Summary by CodeRabbit

* **New Features**
  * Added default and example configurations for MNIST, ResNet, YOLO, and multi-stage workflows.
  * Added cache inspection and cleanup commands.
  * Added automatic local input staging with content-hash deduplication.
  * Improved configuration-relative path resolution and support for multiple model data files.

* **Bug Fixes**
  * Improved Docker output ownership and cache/output directory handling.
  * Prevented inference runs from overwriting unrelated results.

* **Documentation**
  * Updated guidance for caching, staging, Docker usage, and output layouts.

* **Tests**
  * Expanded coverage for caching, staging, Docker execution, and inference outputs.


modelconverter/cli/utils.py
```diff
@@ -101,6 +102,9 @@ def get_configs(
         path_ = resolve_path(path, MISC_DIR)
         if path_.is_dir() or is_nn_archive(path_):
             return process_nn_archive(target, path_, overrides)
+        # Resolve files referenced *inside* the config (calibration data,
+        # scripts, encodings, ...) relative to the config file's directory.
+        set_input_base(path_.parent)
```
klemen1999, line 150: Should this be called at function entry? Otherwise if path is not set we skip the reset of the input base

modelconverter/utils/input_staging.py
```diff
@@ -0,0 +1,383 @@
...
+
+    if src.is_dir():
+        dest = _stage_dir(src, inputs_dir)
+        return _to_container(dest)
+
+    # Config file: stage the whole containing directory so its relative
+    # references resolve alongside it, then stage and rewrite absolute local
+    # references which may live anywhere on the host.
+    if src.suffix.lower() in _CONFIG_EXTS:
+        parent_dest = _stage_dir(src.parent, inputs_dir)
```
klemen1999: This could be very problematic if I for example have config in some big directory because in that case it would try to sync this whole big directory to the cache even though we don't really need any of it.
What if we would go the other way around: Check for files that we need to copy, copy only these files and rewrite the paths to these files to container specific paths for the specific modelconv process?

AI proposed solution:
Conceptually, replace this:

```python
parent_dest = _stage_dir(src.parent, inputs_dir)
config_dest = parent_dest / src.name
```

with:

```python
config_dest = _stage_file(src, inputs_dir)
_rewrite_config_paths(src, config_dest, src.parent, inputs_dir)
```

`_rewrite_config_paths` would:

1. Load the YAML.
2. Walk only known path-bearing fields.
3. Resolve relative local values against `src.parent`.
4. Stage the resolved path using the existing `_stage_value`.
5. Replace that YAML value with the returned `/app/shared_with_container/inputs/...` path.
6. Write the rewritten YAML to the staged config location.

Illustrative shape:

```python
_PATH_FIELDS = {"input_model", "input_bin", "path", "script", "encodings"}

def _rewrite_config_paths(value, *, config_dir, inputs_dir, key=None):
    if isinstance(value, dict):
        return {
            field: _rewrite_config_paths(
                item,
                config_dir=config_dir,
                inputs_dir=inputs_dir,
                key=field,
            )
            for field, item in value.items()
        }

    if isinstance(value, list):
        return [
            _rewrite_config_paths(
                item, config_dir=config_dir, inputs_dir=inputs_dir, key=key
            )
            for item in value
        ]

    if key not in _PATH_FIELDS or not isinstance(value, str):
        return value

    if get_protocol(value) != "file":
        return value

    candidate = Path(value).expanduser()
    if not candidate.is_absolute():
        candidate = config_dir / candidate

    if not candidate.exists():
        return value  # Let normal config validation report the missing path.

    return _stage_value(str(candidate.resolve()), inputs_dir) or value
```

In the real implementation, I would scope `path` specifically to `calibration.path`, rather than treating every YAML field named `path` as a filesystem path. The supported fields should be defined from the config schema and covered by tests for:

- relative model, calibration directory, script, encodings, and OpenVINO `.bin`;
- absolute dependencies outside the config directory;
- remote URLs left unchanged;
- unrelated sibling files never copied.

modelconverter/cli/utils.py
kozlov721: Yes, `set_input_base(None)` is now the first thing `get_configs` does, before the `path is not None` branch, so the base is reset even when no config path is passed.

modelconverter/utils/input_staging.py
```diff
@@ -0,0 +1,1023 @@
...
+    "script": frozenset({"calibration"}),
+    # `RVC4Config.encodings`
+    "encodings": frozenset({"rvc4"}),
+}
+
+# Config fields holding a list of raw tool arguments, mapped to the flags
+# within them whose *value* is a local path. The list as a whole is not a path
+# field: only the token following one of these flags is staged, so the flags
+# themselves and every unrelated argument are passed through untouched.
+_PATH_ARG_FIELDS: dict[str, frozenset[str]] = {
```
klemen1999, line 81: Technically there is `rvc2.mo_args: ["--transformations_config", "custom.json"]` where you can pass in a file (not exposed on Hub but one could technically use this locally)

klemen1999 (commented):
One small change request but overall LGTM, we can merge when this is added

klemen1999 (approved):

PR #263, opened 2026-07-25
Tests refactor

kozlov721:
## Purpose

Previous tests were insufficient. This PR completely refactors the tests, adding coverage reporting and simplified CI.

## Specification

- Adds new `tests/conversion` tests for more reliable testing of model conversions
- More properly tested NN Archive conversion, TFLITE, multi-stage conversion
- **100 %** coverage for the conversion part of `modelconverter`
- `--dev` flag now also installs the `dev` requirements to the image and mounts `tests/` from PWD
  - Makes it possible to run tests with `modelconverter shell rvc4 --dev -c 'pytest -x'`
- benchmarking tests kept unchanged

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

## AI Usage

## Summary by CodeRabbit

* **New Features**
  * Improved image preprocessing for models with different input layouts and channel arrangements.
  * Added support for development container builds with live source and test mounting.
  * Added real-model evaluation coverage for conversion quality and segmentation workflows.

* **Bug Fixes**
  * Fixed RVC2 handling for grayscale and multi-value inputs.
  * Corrected RVC4 encoding archive detection and LLVM runtime compatibility.
  * Improved calibration data handling and ONNX external-data processing.

* **Refactor**
  * Streamlined conversion tests and platform-specific validation across supported targets.
  * Updated publishing and testing workflows for more consistent automation.


.github/workflows/hailo_test.yaml
```diff
@@ -1,38 +0,0 @@
-name: Test - HAILO
-
-on:
-  workflow_dispatch:
-  pull_request:
-    branches: [main]
-    paths:
-      - requirements.txt
```
klemen1999, line 8: Wasn't the point of this so that we don't need to test for example Hailo if changes were made just on for example RVC4 benchmark.py script? Is this optimization not needed anymore?

.github/workflows/publish.yaml
```diff
@@ -1,32 +1,9 @@
-name: Package publishing
+name: Publish images
 
 on:
-  workflow_call:
-    inputs:
```
klemen1999, line 5: We can remove for now but I feel like longterm we'll want to bring this back for cases where e.g. only one package publishing fails. If I understand correctly now we don't have a way to trigger new image build on just one package. Maybe we create a `manual_publish.yaml` for that since probably in this `publish.yaml` it would be messy to add also these inputs. Thoughts?

docker/rvc4/Dockerfile
```diff
@@ -20,8 +20,7 @@ RUN <<EOF
         libcurl4t64     \
         libncurses5     \
         libtinfo5       \
-        llvm-19-runtime \
```
klemen1999, line 23: Why this change? Was it causing issues? It's a bit weird that we are downgrading

modelconverter/packages/base_inferer.py
```diff
@@ -64,12 +65,8 @@ def from_config(
                 else ResizeMethod.RESIZE
                 for inp in config.inputs
             },
-            encoding={
-                inp.name: inp.encoding.to
-                if isinstance(inp.calibration, ImageCalibrationConfig)
-                else Encoding.BGR
-                for inp in config.inputs
-            },
+            encoding={inp.name: inp.encoding.to for inp in config.inputs},
```
klemen1999, line 68: Is now encoding.to always populated so that we don't need BGR default anymore?

modelconverter/__init__.py
```diff
@@ -12,7 +12,7 @@ def load_put_file_plugins() -> None:
     """Registers any external put file plugins."""
     eps = entry_points()
     put_file_plugins = eps.select(group="put_file_plugins")
-    for entry_point in put_file_plugins:
+    for entry_point in put_file_plugins:  # pragma: no cover
         plugin_class = entry_point.load()
         PUT_FILE_REGISTRY.register_module(module=plugin_class)
```
klemen1999: We should probably switch to using `.register` since `.register_module` is deprecated

.pre-commit-config.yaml
```diff
@@ -2,21 +2,14 @@ default_language_version:
...
-    rev: v0.15.16
+    rev: v0.16.0
     hooks:
       - id: ruff
         args: [--fix, --exit-non-zero-on-fix]
         types_or: [python, pyi, jupyter]
       - id: ruff-format
         types_or: [python, pyi, jupyter]
 
-  - repo: https://github.com/PyCQA/docformatter
```
klemen1999, line 13: Is this preparation for switching away from epytext docstrings? What is the current state? Would the CI fail if this was still in? Are we in the "in-between" stage where both docstrin styles are used?

klemen1999 (commented):
Left some comments but overall LGTM and close to being mergable.
Overall I focused more on the changes in the core modules rather than test cases since I assume those should be good.

requirements-eval.txt
klemen1999, line 1: What is this for? Seems like something left from local testing no?

.pre-commit-config.yaml
kozlov721: Yeah this is a sort of "in-between" stage. `docformatter` have some conflicting rules with `ruff` on top of that so I just removed it already.

docker/rvc4/Dockerfile
kozlov721: Yeah the `tflite` conversion did not work because of shared libraries.

requirements-eval.txt
kozlov721: These are special requirements used inside the `dev` docker images so we can install `luxonis-eval` to use the metrics from there for proper evaluation of converted models.

.github/workflows/hailo_test.yaml
kozlov721: This is now done more dynamically `.github/scripts/changed_platforms.sh`

.github/workflows/publish.yaml
kozlov721: Added back platform and also tool-version inputs.

requirements-eval.txt
klemen1999: So luxonis-eval is actively used now in modelconverter testing? Because I didn't see it but maybe I missed it.
And If so why this use this specific commit?

kozlov721: Yes `luxonis-eval` is one of `modelconverter`'s test dependencies. We used to have the metrics just copied here so I figured it would be better to use `luxonis-eval` directly.

`luxonis-eval` is not released on PyPI and has only one older tag so we have to point to some other specific commit, which is the current `main`. 

I'd need to look into it again but I think some tool had an issue with using just regular `@main`.

kozlov721: Ideally we should also make a new tag on the `luxonis-eval` repo, even though we should use a hash for security reasons anyway. Pointing to a tag is still going to be better than to a random commit on main.

klemen1999: I see it now yeah. Ok I think if what is here currently works then it makes sense to leave it as is and handle better luxonis-eval dependency in separate PR. This is because we are introducing a lot of breaking changes in the next release of luxonis-eval so refactor here will be needed anyway.

modelconverter/packages/base_inferer.py
kozlov721: Yes — `to` is non-optional with a `BGR` default, and `_validate_encoding` normalizes the rest, so it's always populated.

klemen1999 (approved):

PR #270, opened 2026-08-07
Fix HIL benchmark test dependencies

tjb7:
fixes failing HIL BOM tests by adding the hypothesis framework

## Summary by CodeRabbit

* **Tests**
  * Updated the hardware-in-the-loop test setup to install a compatible Hypothesis version before running tests.

### DIFF
--- a/tests/test_benchmark/run_hil_tests.sh
+++ b/tests/test_benchmark/run_hil_tests.sh
@@ -25,8 +25,7 @@ python3 -m venv venv
 source venv/bin/activate
 
 # Install dependencies
-pip install -r requirements.txt
-pip install pytest
+pip install ".[dev]"
 
 pip install hil-framework --upgrade \
   --index-url "https://__token__:$PAT_TOKEN@gitlab.luxonis.com/api/v4/projects/213/packages/pypi/simple" \


tests/test_benchmark/run_hil_tests.sh
```diff
@@ -27,6 +27,7 @@ source venv/bin/activate
 # Install dependencies
 pip install -r requirements.txt
 pip install pytest
+pip install "hypothesis~=6.163"
```
tjb7: not sure if this is actually the right place for this. Should tests just install all dev dependencies instead?

klemen1999: Would a cleaner fix all around for you be:
`pip install .[dev]`
instead of manually installing requirementes.txt, pytest and hypothesis? This `[dev]` marker will install `requirements-dev.txt` and `requirements-bench.txt` which is likely all you need?

klemen1999: Yes I think so, try the suggestion I made in https://github.com/luxonis/modelconverter/pull/270#discussion_r3734255094

klemen1999 (approved):

PR #271, opened 2026-08-07
Fix `modelconverter analyze` command

dtronmans:
## Purpose
Order of fixes performed:

1) The pre-created debug output directory that `snpe-net-run --debug` would use was "predicted" incorrectly. Solution: parse SNPE's missing-utput error, create the parent directory on-device then retry.
2) `--output-dir` should have been passed to `snpe-net-run` (in the other RVC4 paths it is)
3) Changed the analyze cycle to be sample-by-sample instead of one large debug pass because dumping every intermediate tensor for 256 all at once was too heavy

- Added `--image subset <num-images>` parser argument to optionally limit the number of images used in the analyze loop.

Working usage now:

`modelconverter analyze --dlc-model-path <dlc-model-path>.dlc --onnx-model-path <onnx-intermediate-path>.onnx --image-dirs <flat-dir-of-images> --device-ip <device-ip> --image-subset N`

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

## AI Usage

## Summary by CodeRabbit

* **New Features**
  * Added an optional image limit for analysis runs.
  * Added support for PNG, JPG, JPEG, and NPY input files.
  * Added per-sample progress reporting and comparison statistics.
  * Analysis results can now be exported as CSV files.
  * Added configurable raw-input preparation and inference logging.

* **Bug Fixes**
  * Improved handling of empty, invalid, or mismatched input sets.
  * Improved inference reliability by recovering from missing output directories.
  * Added clearer validation when an invalid image limit is provided.

* **Documentation**
  * Clarified which DLC files to select for analysis.


modelconverter/packages/rvc4/analyze.py
```diff
@@ -14,23 +14,28 @@
...
         device_ip: str | None,
         device_id: str | None,
         dlc_model_path: str,
         image_dirs: dict[str, str],
+        image_subset: int | None = None,
     ):
         super().__init__(dlc_model_path, image_dirs)
         _, device_adb_id = get_device_info(device_ip, device_id)
         self.handler = create_handler(device_ip, device_adb_id)
+        self.image_subset = image_subset
```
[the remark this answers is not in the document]
dtronmans: True, fixed in [2e8031d](https://github.com/luxonis/modelconverter/pull/271/commits/2e8031dd91663d66e415a8ecb248ecc323449579)

modelconverter/packages/rvc4/analyze.py
```diff
@@ -69,12 +111,25 @@ def _resize_image(
 
     def _prepare_input_matcher(self) -> dict[str, dict[str, str]]:
         image_names = {
-            k: sorted(Path(v).glob("*")) for k, v in self.image_dirs.items()
+            k: sorted(
+                path
+                for path in Path(v).glob("*")
+                if path.suffix.lower() in self.SUPPORTED_INPUT_SUFFIXES
+            )
```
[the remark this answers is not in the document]
dtronmans: Sorry but will leave this one out, at least in my case I have not encountered a directory that incidentally ends with these extensions yet (and these are not artifacts of any of our repositories), this is more of a user error

modelconverter/packages/rvc4/analyze.py
```diff
@@ -168,14 +238,70 @@ def _prepare_output_dirs(self, *, debug: bool) -> None:
...
+            seen_dirs.add(parent_dir)
+            missing_dirs.append(parent_dir)
+
+        return missing_dirs
+
+    def _execute_dlc_command(
+        self, command: str, *, max_missing_output_dir_retries: int = 3
+    ) -> None:
+        work_dir = f"/data/modelconverter/{self.model_name}"
+        for attempt in range(max_missing_output_dir_retries + 1):
```
klemen1999, line 290: Why multiple retries? I assume SNPE stderr includes all the missing output dirs if it fails. In that case after first failure we have exact issues, we can fix it and inside "except" already handle and perform the correct no?

klemen1999 (commented):
I'd propose a better flow overall. If I understand the code as is now correctly we are doing:
```
For each sample:
   try to guess the output structure
   if it fails parse through SNPE errors and create missing folders
   run snpe-net-run to create outputs
   copy outputs to host
   delete everything
```
But since we are dealing with the same model and we are just changin the inputs between samples I think a better flow would be:
```
one-time setup:
    push DLC
    create the dir structure on the device including output directories (use SNPE with a dummy/first input directly, don't need to even try to guess it with ONNX)

for each sample:
    replace staged input files and input_list.txt
    delete old output .raw files, but keep their parent directories
    run SNPE
    pull outputs to the host
    compare them locally
    delete local temporary output files
```
This way we are technically doing one more inference (the first dummy one) but we don't need to keep trying and guessing the structure for every sample.
Does this make sense?

klemen1999:
For CI fixes we need to merge in `main` branch

modelconverter/packages/rvc4/analyze.py
```diff
@@ -67,15 +156,37 @@ def _resize_image(
...
         return image.astype(np.uint8)
 
-    def _prepare_input_matcher(self) -> dict[str, dict[str, str]]:
+    def _get_selected_image_paths(self) -> dict[str, list[Path]]:
         image_names = {
-            k: sorted(Path(v).glob("*")) for k, v in self.image_dirs.items()
+            k: sorted(
+                path
+                for path in Path(v).glob("*")
+                if path.suffix.lower() in self.SUPPORTED_INPUT_SUFFIXES
```
klemen1999: Let's add path.is_file() check as well so we don't try to read directories named e.g. sample.npy

modelconverter/packages/rvc4/analyze.py
```diff
@@ -301,6 +450,11 @@ def _compare_to_onnx(
                 )
                 statistics.append([layer_name, *layer_stats])
 
+        return statistics
+
+    def _write_layer_comparison_csv(
+        self, statistics: list[list], layer_names: list[str]
+    ) -> None:
         output_dir = f"{constants.OUTPUTS_DIR!s}/analysis/{self.model_name}"
```
klemen1999: We should do:
```
output_dir = constants.OUTPUTS_DIR / "analysis" / self.model_name
output_dir.mkdir(parents=True, exist_ok=True)
```
since analysis/model_name is not created anywhere (if you start fresh)

klemen1999 (commented):
Generally LGTM, just two smaller mistakes

modelconverter/packages/rvc4/analyze.py
dtronmans: Thanks, addressed both in 94f5b2c

modelconverter/packages/rvc4/analyze.py
dtronmans: This is actually something similar that CodeRabbit suggested that I discarded in [this thread](https://github.com/luxonis/modelconverter/pull/271#discussion_r3734960756), just because a directory named *.npy or *.png isn't really something we return in any of our repositories. But it's true that this is something that could potentially happen so I will implement the suggestion

klemen1999: Yeah it s a very far edge case. But since it is just one function call and easy addition we can just add it

dtronmans: Addressed in [172b694](https://github.com/luxonis/modelconverter/pull/271/commits/172b694969b944d6884ce33f6e180702c6b80e0a)

modelconverter/packages/rvc4/analyze.py
dtronmans: Addressed in [172b694](https://github.com/luxonis/modelconverter/pull/271/commits/172b694969b944d6884ce33f6e180702c6b80e0a)

klemen1999 (approved):

PR #274, opened 2026-08-13
Migrate the package docs to Google docstyle, document the public API, and gate CI on both

kozlov721:
## Purpose
Our in-code documentation is mid-migration to Google docstyle. `luxonis-ml` has
moved; `modelconverter` was still on Epytext, and its CLI commands used a third
style (numpydoc) on top of that. This brings `modelconverter` in line with
`luxonis-ml`'s `3460597` ("Enhance documentation"), documents the public API,
and puts both under CI so neither can regress.

The docs also could not be built: `pydoctor` emitted 14 warnings, so a
warnings-as-errors build was impossible.

## Specification
**Docstyle migration.** `docformatter` style `epytext` → `google`, ruff's `D`
rules enabled with the same ignore list `luxonis-ml` uses. Every Epytext
docstring converted (`@type`/`@param`/`@rtype`/`@return`/`@raises` →
`Args:`/`Returns:`/`Yields:`/`Raises:`; `C{}`/`L{}` → literals/references). The
cyclopts CLI commands' numpydoc `Parameters` blocks are converted too — these
render into `modelconverter --help`.

**Warning-free docs build.** `pydoctor --docformat=google -W` now exits 0. Two
causes: single backticks are cross-references in Google/RST mode, so
unresolvable names had to become literals; and pydoctor `ast.parse`s every
string literal inside an annotation, so `Parameter(help="some help")` and
`Field(alias="from")` failed to parse. `launcher`'s six option descriptions
moved from `Parameter(help=...)` into its (previously absent) docstring, which
cyclopts reads identically; `EncodingConfig.from_` now uses a plain `Field()`
default. Those 7 warnings predate this PR.

**Public API documented and enforced.** Every public module, package, class,
method and function now has a docstring, and `D100`–`D107` are enforced.
`tests/*` is exempt from `D103` alone — a docstring on each of the 410
`def test_*` would be noise, while the module, class and helper docstrings
around them are not.

**Input formats corrected.** The converter takes ONNX, OpenVINO IR and TFLite,
and the accepted set is per-target: the CLI rejects an IR for anything but
RVC2/RVC3, TFLite reaches RVC4/Hailo via `snpe-tflite-to-dlc` and
`translate_tf_model`, and RVC2/RVC3 via an ONNX conversion their exporter logs
as experimental. Each sub-package states what it accepts; the top-level defers
to them. `.pt`/`.pth` are not documented, being no longer relevant.

**CI.** A `pre-commit` job (running `prek`, as `luxonis-train` and
`luxonis-ml`'s `dev/uv` branch do) and a `docs` job (`pydoctor -W`) gate every
test job: `pre-commit → docs → unit → integration → coverage`.

No runtime dependency changes. `requirements-dev.txt` swaps `pre-commit~=4.6`
for `prek~=0.4.13`; `.pre-commit-config.yaml` is unchanged, as prek reads it
as-is. CI gains a `pydoctor~=25.10` install in the docs job (no project deps
needed — pydoctor reads sources rather than importing them).

No behaviour change. Every docstring commit is AST-proven docstring-only
against its parent: no signature, logic, import or behaviour changes. Two
deliberate exceptions, both called out in their commit messages:

- `EncodingConfig.from_` declares its alias via a plain `Field()` default
  instead of `Annotated[..., Field(...)]`. Identical pydantic semantics —
  alias parsing, serialization alias and default all verified unchanged.
- One line of `rvc4/analyze.py`: a `return` inside a `try` moved into an `else`
  block. That `TRY300` has been failing on `main`; documenting the file brought
  it under the hook and blocked the commit. All three paths are unchanged.

Reviewers should note that ruff derives publicness from `__all__`, so
`D101`/`D102` never fired for `Layout`, `QuantizationMode`, `InputFileType` and
`InputFileType.from_path` in `utils.types`. They are documented here, but the
newly enabled rules will not catch the next such symbol.

## Testing & Validation
- `pydoctor --docformat=google -W` exits 0 (verified in a venv containing only
  pydoctor, i.e. exactly what the CI job does), and verified to exit non-zero
  on a deliberately broken reference.
- `ruff check .` — all checks passed, no exclusions. `ruff format --check .` —
  116 files clean.
- `prek run --all-files` — 16 hooks pass; verified to exit 1 on a bad file.
- 827 unit tests pass, 1 pre-existing skip.
- `modelconverter --help` diffed before/after for all seven commands: every
  parameter still renders; the only changes are imperative summaries, a `fron`
  typo fix, and quoted literals now rendering as code.
- The ~300 new docstrings were written by a multi-agent pass and then
  adversarially verified against the implementation, which corrected 52 factual
  errors — including `EncodingConfig`'s `from_`/`to` documented backwards, a
  `Raises:` that cannot fire, and an undocumented in-place mutation of the
  caller's config.
- Conversion tests (`tests/conversion`) were not run — they need the per-backend
  Docker images and, for some, an attached device.

## Deployment Plan
None / not applicable — documentation, lint configuration and CI only. No
release or rollout step; reverting the branch fully restores the previous state.

## AI Usage
Assisted-by: Claude Code:claude-opus-5 ruff pydoctor prek pytest

Submitted code was reviewed by a human: NO

## Summary by CodeRabbit

* **Documentation**
  * Expanded and standardized developer and API documentation across the project.
  * Clarified supported model formats and inference inputs, including OpenVINO IR, TensorFlow Lite, images, NumPy arrays, and raw buffers.
  * Added guidance for LDF calibration datasets and multi-input limitations.
  * Improved command-line help and configuration guidance.
* **Chores**
  * Added pre-commit validation and API documentation builds to CI.
  * Updated linting and development validation tooling.
  * Added generated documentation files to ignore rules.


modelconverter/__main__.py
```diff
@@ -1,3 +1,14 @@
...
+
+Defines the commands the CLI exposes: converting a model for a platform,
+running inference, benchmarking and analyzing it on a device,
+visualizing the analysis, packaging a model into an NN Archive, opening
+a shell in a platform's container and managing the cache. The commands
+that need the vendor conversion tools -- ``convert``, ``infer`` and
+``shell`` -- are re-run inside the Docker image of the requested
+platform unless they already run inside one; the others run on the host.
+"""
+
```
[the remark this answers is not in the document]
kozlov721: The routing claim does not hold. `benchmark`, `analyze`, `visualize`, `archive` and the cache commands are registered with `@app.meta.command`, so cyclopts dispatches them directly and `launcher()` never sees them. Only the inner-app commands `convert`, `infer` and `shell` reach `launcher()`, and each of them requires `platform`. The docstring describes this behavior correctly, so it stays as it is.

modelconverter/platforms/rvc2/__init__.py
```diff
@@ -0,0 +1,13 @@
+"""Conversion backend for the RVC2 platform.
+
+Holds the exporter, inferer and benchmark for RVC2. The exporter
+accepts an ONNX model, an OpenVINO IR pair or a TFLite model; a TFLite
+model is converted to ONNX first, which the exporter itself logs as
+experimental. The conversion then goes through OpenVINO: its model
+optimizer turns the ONNX model into an OpenVINO IR and ``compile_tool``
+compiles that into a ``.blob``.
```
klemen1999: Let's mention superblob here as well

.github/workflows/ci.yaml
```diff
@@ -75,8 +75,73 @@ env:
...
+        run: python -m pip install 'prek~=0.4.13'
+
+      - name: Run hooks
+        run: prek run --all-files --show-diff-on-failure --color=always
+        env:
+          # CI can run from a checked-out protected branch via `workflow_call`.
+          # Keep the hook for local commits, but skip it for a full lint run.
+          SKIP: no-commit-to-branch
+
+  docs:
```
klemen1999, line 109: Just a note: We'll probably replace this with the Luxonis Docs script right? Same as we have in luxonis-ml currently

klemen1999 (approved):
Generally LGTM just one small nitpick

.github/workflows/ci.yaml
kozlov721: We no longer have any script in luxonis-ml for building the docs. The script was only needed for the gh pages.

klemen1999: I meant [this](https://github.com/luxonis/luxonis-ml/blob/faab695545ec0031cba036b3d834f5c7b0e11b68/.github/workflows/export-pydoctor-reference.yaml#L60)

kozlov721: Oh yeah, yes I expected to create an equivalent workflow here as well when it's all setup on the docs side (not sure what is the state of things there).

PR #275, opened 2026-08-13
Add INT16_STANDARD quantization mode for RVC4

rolandocortez:
## Summary

Adds a new `INT16_STANDARD` quantization mode for RVC4, using 16-bit weights and 16-bit activations.

RVC4 enables SNPE per-channel quantization by default because it generally improves accuracy for the existing quantization modes. During W16A16 validation, however, enabling per-channel quantization caused severe target-side degradation, while the native W16/A16 path without it remained healthy.

`INT16_STANDARD` therefore uses the native SNPE W16/A16 path without applying the default per-channel flag.

## Changes

- Add `INT16_STANDARD` to `QuantizationMode`
- Configure RVC4 quantization with:
  - `--weights_bitwidth 16`
  - `--act_bitwidth 16`
- Do not apply the default `--use_per_channel_quantization` flag for `INT16_STANDARD`
- Report `int16` precision in generated NNArchives
- Add unit and conversion coverage for the new preset
- Add `INT16_STANDARD` to the README and defaults documentation
- Add the missing `INT8_INT16_MIXED_ACCURACY_FOCUSED` entry to the `defaults.yaml` available-modes comment

## Validation

Evaluated the requested quantization modes across four model types using LuxonisEval, with RVC4 throughput measured using `modelconverter benchmark`.

### Accuracy

| Model | Metric | ONNX | FP16 | INT8 | INT16_STANDARD |
| --- | --- | ---: | ---: | ---: | ---: |
| Detection | BBox AP | 0.7466 | 0.7377 | 0.6976 | 0.7086 |
| Instance Segmentation | BBox AP | 0.8173 | 0.8192 | 0.7778 | 0.8005 |
| Instance Segmentation | Mask AP | 0.7089 | 0.7090 | 0.6637 | 0.6928 |
| Keypoints | BBox AP | 0.9386 | 0.9414 | 0.9283 | 0.9364 |
| Keypoints | Keypoint mAP | 0.6561 | 0.6458 | 0.5935 | 0.6545 |
| Semantic Segmentation | Jaccard | 0.7391 | 0.7362 | 0.4524 | 0.7697 |
| Semantic Segmentation | F1 | 0.8500 | 0.8480 | 0.6230 | 0.8699 |

Across these four tested model families, `INT16_STANDARD` consistently preserves task quality better than `INT8_STANDARD` on the representative metrics.

### RVC4 benchmark

| Model | FP16 FPS | INT8 FPS | INT16_STANDARD FPS |
| --- | ---: | ---: | ---: |
| Detection | 793.75 | 689.34 | 714.49 |
| Instance Segmentation | 520.79 | 633.36 | 477.45 |
| Keypoints | 676.78 | 695.81 | 636.60 |
| Semantic Segmentation | 35.19 | 165.51 | 51.76 |

The throughput trade-off is model-dependent. The main demonstrated benefit of `INT16_STANDARD` in this matrix is therefore improved quality preservation compared with INT8, rather than a consistent FPS advantage.

### INT16_ACC follow-up

I also evaluated the proposed `INT16_ACC` variant from the review discussion. The literal enhanced-parameter + enhanced-activation configuration did not show a consistent accuracy benefit across the tested models, so it is intentionally left out of this PR for now.

## AI Usage

Assisted-by: ChatGPT

## Summary by CodeRabbit

- **New Features**
  - Added RVC4 `INT16_STANDARD` quantization support for 16-bit weights and activations.
  - Added documentation and configuration support for available RVC4 quantization modes.
  - Updated converted archive precision handling for FP16, FP32, INT8, and INT16 modes.

- **Bug Fixes**
  - Improved calibration behavior for INT16 quantization while preserving existing INT8 and per-row quantization options.

- **Tests**
  - Added coverage for INT16 conversion, archive precision, calibration, and quantization settings.


modelconverter/packages/rvc4/exporter.py
```diff
@@ -159,7 +159,10 @@ def calibrate(self, dlc_path: Path) -> Path:
         self._add_args(args, ["--input_dlc", dlc_path])
         self._add_args(args, ["--output_dlc", quantized_dlc_path])
 
-        if self.use_per_channel_quantization:
+        if (
+            self.use_per_channel_quantization
+            and self.quantization_mode != QuantizationMode.INT16_STD
```
klemen1999, line 164: Let's add a small comment here that shortly explains why INT16_STD is a special case here

klemen1999 (commented):
Overall looks good but before merge I'd like to have a bit more results for some other types of models. The current test in the description is a bit fishy since W16A18 is same as FP16. I'd like to have some more comparisons.
And let's include `INT16_ACC` (which would be defined same as INT6_STD + `self._add_args(args, ["--param_quantizer", "enhanced"])` and `self._add_args(args, ["--act_quantizer", "enhanced"])`) and test with this one same models as well. And ideally also include FPS (`modelconverter benchmark` ..) into the comparison list together with the accuracy.
Main goal is to have this backed by some data where we can see that it is worth adding.

Note: You should be able to use agents to make a full automated script since this is repeating the same things across different QuantizationModes. Just noting so you don't go doing this too manually if there is no need.

rolandocortez:
> Overall looks good but before merge I'd like to have a bit more results for some other types of models. The current test in the description is a bit fishy since W16A18 is same as FP16. I'd like to have some more comparisons. And let's include `INT16_ACC` (which would be defined same as INT6_STD + `self._add_args(args, ["--param_quantizer", "enhanced"])` and `self._add_args(args, ["--act_quantizer", "enhanced"])`) and test with this one same models as well. And ideally also include FPS (`modelconverter benchmark` ..) into the comparison list together with the accuracy. Main goal is to have this backed by some data where we can see that it is worth adding.
> 
> Note: You should be able to use agents to make a full automated script since this is repeating the same things across different QuantizationModes. Just noting so you don't go doing this too manually if there is no need.

Addressed The consolidated results are now in the PR description. I also evaluated INT16_ACC across the same models. Based on the results, I’m leaving it out of the PR for now

modelconverter/packages/rvc4/exporter.py
rolandocortez: Addressed in [e95fe50](https://github.com/luxonis/modelconverter/pull/275/commits/e95fe50ca28d91e8fcffb7bf240f43e828348ac6)

klemen1999 (approved):

PR #288, opened 2026-08-28
Add strict validation for RVC4 quantization overrides

rolandocortez:
## Summary

- Add opt-in `rvc4.strict_quantization_overrides` validation, disabled by default.
- Validate custom quantization override names against separate activation and parameter namespaces in the effective ONNX model before invoking SNPE.
- Reject nonexistent names and names placed in the wrong encoding group.
- Keep ONNX `Constant` outputs in the parameter namespace for this check.
- Normalize both `--quantization_overrides PATH` and `--quantization_overrides=PATH` in `RVC4Config` before they reach the exporter.
- Reject missing, multiple, or conflicting quantization override sources during config validation.
- Keep validation opt-in, so existing conversion behavior is unchanged when strict mode is disabled.

SNPE 2.41 is fairly permissive when processing quantization overrides. A parameter placed under `activation_encodings` can be recovered and applied, while an activation placed under `param_encodings` can still convert successfully but have its override silently ignored.

Because of that, strict mode keeps the two encoding groups separate instead of relying on SNPE to recover misplaced entries.

Partial encoding files are still valid, since strict mode only validates the names that are actually provided.

QAIRT-specific restrictions on individual parameters are left out of scope for this check. Strict validation is currently ONNX-only when effective overrides are present. FP16 conversions are unaffected because quantization overrides are not consumed in that path.

## Testing

- `25 passed` - strict quantization override unit tests
- `9 passed` - RVC4 config normalization tests
- `956 passed, 5 skipped` - full unit test suite
- `4 passed` - RVC4/SNPE 2.41 integration tests

## Summary by CodeRabbit

* **New Features**
  * Added optional strict validation for RVC4 quantization overrides.
  * When enabled, validates configured activation and parameter encoding names against tensors in the effective ONNX model.
  * Supports tensor names from inputs, outputs, initializers, node connections, sparse initializers, and external-data models.

* **Bug Fixes**
  * Invalid override names are reported before external conversion tools run.
  * Strict validation is limited to ONNX input models and configured encoding sources.


modelconverter/utils/encodings.py
```diff
@@ -128,3 +131,52 @@ def parse_encodings(value: Any) -> "Encodings":
...
+        for sparse in graph.sparse_initializer
+        if sparse.values.name
+    )
+
+    activation_names = set()
+    activation_names.update(_non_empty_names(graph.input))
+    activation_names.update(_non_empty_names(graph.output))
+    for node in graph.node:
+        activation_names.update(name for name in node.input if name)
+        activation_names.update(name for name in node.output if name)
```
N950: Thanks Rolando
I think this will result in op_type `Constant` being treated as an activation, but constants are parameters in encodings/SNPE context (e.g. B param in a `Mul` op), while in onnx constants are the outputs of op_type `Constant` nodes.

There is a chance that might get hidden when using a simplified onnx because iirc onnxsim can auto converts constants to initializers

modelconverter/utils/encodings.py
```diff
@@ -128,3 +131,52 @@ def parse_encodings(value: Any) -> "Encodings":
...
+
+    activation_names.difference_update(parameter_names)
```
N950: Worth thinking what issues this could silently hide, ideally our logic produces activations/parameters that are disjoint and maybe even asserts that they are so that it points to any wrong assumptions.

rolandocortez: Yep, good catch. I checked this a bit more with SNPE and it’s actually looser here than I expected, so I changed the check to only verify that the name exists in the final model

modelconverter/platforms/rvc4/exporter.py
```diff
@@ -289,6 +302,93 @@ def _generate_io_encodings(self, encodings: Encodings) -> Path:
...
+                description=f"raw --quantization_overrides {path}",
+                path=path,
+            )
+            for path in self._raw_quantization_override_paths(
+                self._snpe_onnx_to_dlc
+            )
+        )
+        return sources
+
+    def _validate_quantization_overrides(self) -> None:
```
klemen1999: General rule: If function is private let's put it after the public functions and especially after public functions that call it. So in this case I'd put this all the way to the bottom

modelconverter/utils/encodings.py
```diff
@@ -139,3 +141,50 @@ def parse_encodings(value: "ParamValue | Encodings") -> "Encodings":
             ),
         }
     )
+
+
+def _non_empty_names(values: Any) -> set[str]:
+    return {value.name for value in values if value.name}
+
+
+def collect_onnx_tensor_names(
```
klemen1999: Nitpick: I prefer to have functions that are called below the ones that are calling it. So in this case I'd prefer to have first the `validate_quantization_override_names`, then `collect_onnx_tensor_names` and last `_non_empty_names`

modelconverter/platforms/rvc4/exporter.py
klemen1999, line 1: I think both `_raw_quantization_override_paths` and `_quantization_override_sources` shouldn't be needed. We should normalize this already at the RVC4Config level and to the exporter only one encoding path should come. So at the config level we check `snpe_onnx_to_dlc_args` and then:
- extract `--quantization_overrides PATH` or `--quantization_overrides=PATH`
- reject zero/multiple/missing values
- reject it when `rvc4.encodings` is also set
- remove the raw flag from `snpe_onnx_to_dlc_args`

And then logic in exporter should essentially just be:
```python
if self._strict_quantization_overrides and self._encodings is not None:
    validate_quantization_override_names(
        self._encodings, self._input_model
    )
```

klemen1999 (commented):
Left some comments for improvements

modelconverter/platforms/rvc4/exporter.py
rolandocortez: Addressed in [e5ffdb0](https://github.com/luxonis/modelconverter/pull/288/commits/e5ffdb03855daf2c44cc3f51c7aa691cd560c8d0)

modelconverter/utils/encodings.py
rolandocortez: Addressed in [e5ffdb0](https://github.com/luxonis/modelconverter/pull/288/commits/e5ffdb03855daf2c44cc3f51c7aa691cd560c8d0)

modelconverter/platforms/rvc4/exporter.py
rolandocortez: Yeah, good catch. Moved the raw override normalization to `RVC4Config`, so the exporter has a narrower scoped responsability

klemen1999 (approved):

modelconverter/utils/encodings.py
```diff
@@ -175,3 +177,50 @@ def parse_encodings(value: "ParamValue | Encodings") -> "Encodings":
...
+    )
+    for node in graph.node:
+        names.update(name for name in node.input if name)
+        names.update(name for name in node.output if name)
+
+    return names
+
+
+def _non_empty_names(values: Any) -> set[str]:
+    return {value.name for value in values if value.name}
```
N950, line 243: This is checking that the name of an activation/parameter exists in the onnx ? if an encoding json wrongly lists a parameter under activations or the opposite will it pass strict_quantization_overrides ?

rolandocortez: Yep, it will. I checked this with SNPE 2.41 and it's actually fine with the same existing tensor being provided from either group. I even tested an initializer that ModelConverter classifies as a parameter only under activation_encodings, and SNPE applied it.

So I kept the strict check to name existence only, otherwise it would reject cases SNPE itself accepts

N950: iirc only a parameter inside activation list will pass snpe onnx to dlc, an activation inside parameters list will not work and will be silently ignored in snpe onnx to dlc 
even if both pass in 2.41.0 because snpe is treating the json in a non strict way, modelconverter should stick to the defined encoding act/param structure without assuming that snpe onnx to dlc will soft recover or not

rolandocortez: Yeah, you were right. Thanks for catching this. Addressed in [38ccfb9](https://github.com/luxonis/modelconverter/pull/288/commits/38ccfb9be15941dc449b0dfb74057fc4b262367b)

kozlov721 (approved):

N950 (approved):

PR #294, opened 2026-09-02
Manage the host environment with uv

kozlov721:
## Purpose

luxonis-ml and luxonis-train manage their environments with uv. modelconverter still resolves loose requirements files with pip, so every CI run and every contributor install can pick different versions. This PR brings the host environment and CI in line with the sibling repositories. A follow-up PR moves the Docker images.

## Specification

- The dependency metadata moves into `pyproject.toml`. The `bench` and `analysis` extras stay public extras. The `dev` extra becomes the `dev` dependency group, with `lint` and `docs` split out, as in luxonis-ml.
- `uv.lock` pins the host environment. `.python-version` pins the interpreter (and leaves `.gitignore`, which ignored it).
- A lock constraint holds onnxruntime below 1.24 on Python 3.10. Newer releases declare 3.10 support but ship no cp310 wheels.
- An override drops `opencv-python-headless` from the resolution, as in luxonis-ml. albumentations pulls it in next to `opencv-python`, and both ship the same `cv2` module.
- CI installs with `uv sync --locked` and runs tools with `uv run`. The pyright action finds the locked pyright on `PATH`.
- `python-publish.yml` builds with `uv build`.
- The HIL script installs the `dev` group with uv, because the `dev` extra no longer exists.
- A monthly workflow runs `uv lock --upgrade` and opens a PR, as in luxonis-ml.
- The Docker image builds still read `requirements.txt` and `requirements-dev.txt`. Both files stay, with a header that marks them as image-only, until the follow-up PR moves the images to uv.
- `requirements-typecheck.txt` and `requirements-eval.txt` stay as files on purpose. Both install with `--no-deps` and carry pins no resolver accepts.

- No runtime dependency additions. The published metadata lists the same dependencies as before, minus a duplicate `onnxsim` entry.
- The published package loses the `dev` extra. `pip install modelconv[dev]` stops installing the dev tools; contributors use `uv sync` instead. The `bench` and `analysis` extras are unchanged.

## Deployment Plan

None / not applicable.

## Testing & Validation

- `uv run --no-sync pytest tests/unit -n auto`: 935 passed, 6 skipped (the usual toolchain-gated skips).
- pyright over the synced environment, with luxonis-ml from `main` and `requirements-typecheck.txt` installed: only the 4 `hailo_sdk_client` import errors remain. The Hailo wheel resolves them in CI; a local gcloud re-auth was not possible.
- `uv run --locked --only-group lint prek run --all-files` and `uv run --locked --only-group docs python -m pydoctor ...`: both pass with the exact commands CI now runs.
- `uv build` produces the sdist and the wheel. The wheel metadata matches the old dynamic metadata.

## AI Usage

Assisted-by: Claude Code:claude-fable-5 uv pyright

Submitted code was reviewed by a human: NO

The author is taking the responsibility for the contribution: YES

🤖 Generated with [Claude Code](https://claude.com/claude-code)


kozlov721:
Superseded by `chore/uv-lockfiles`, which locks the four image environments as well as the host.

PR #311, opened 2026-09-15
Preserve raw RVC4 quantization overrides

rolandocortez:
## Summary

`rvc4.encodings` is a convenience layer over SNPE's `--quantization_overrides`, but it was previously parsed into ModelConverter's `Encodings` model before being passed to SNPE.

That parsing can be lossy when an encoding file contains valid SNPE fields that ModelConverter does not know about yet.

This PR preserves raw quantization override data instead and leaves SNPE as the authority on the vendor schema.

## Changes

- Preserve `rvc4.encodings` supplied as a path, inline JSON, or dictionary without coercing it through `Encodings`.
- Treat direct `--quantization_overrides` arguments the same way as `rvc4.encodings`.
- When `normalize_io_encodings=False`, pass file-based overrides directly to SNPE without reserializing them.
- Keep `strict_quantization_overrides` as an opt-in name check without transforming the payload.
- Keep `normalize_io_encodings` behavior for exposed model IO, while preserving unrelated and unknown override fields.
- Keep compatibility with callers that already provide an `Encodings` instance.
- Resolve relative direct `--quantization_overrides` paths consistently with `rvc4.encodings`.

## Validation

- 339 relevant host/unit tests pass with the current host dependencies.
- 5 RVC4 encoding conversion tests pass with QAIRT/SNPE 2.41.

## Summary by CodeRabbit

* **New Features**
  * RVC4 quantization overrides now support raw JSON mappings, files, and typed encoding configurations.
  * Override payloads pass through unchanged when IO normalization is disabled.
  * Optional IO normalization updates exposed input and output encodings while preserving other override data.
  * Quantization override names are validated for raw and typed configurations.
  * Quantization override file paths can be provided directly through configuration or command-line options.

* **Documentation**
  * Updated configuration guidance to describe raw JSON pass-through and IO normalization behavior.


modelconverter/utils/encodings.py
```diff
@@ -186,16 +188,68 @@ class _ONNXEncodingNames(NamedTuple):
     parameter_names: set[str]
 
 
+class _QuantizationOverrideNames(NamedTuple):
```
kozlov721: I'd say the custom named tuple just for two sets is an overkill. We can use regular tuple and just unpack it right away when calling this function.

```python
activations, parameters = collect_quantization_override_names(...)
```

kozlov721 (approved):
One nitpick but otherwise LGTM

modelconverter/utils/encodings.py
rolandocortez: Oh yeah, that makes sense. I mirrored the ONNX helper for clarity, but you are right, with a single caller it’s overkill. Changed in [1f14add](https://github.com/luxonis/modelconverter/pull/311/commits/1f14add45d1a15afa5be3cc684f17c14e467265f), thanks!

PR #312, opened 2026-09-16
Keep monitored RVC4 DAI benchmarks on one device

rolandocortez:
## Summary

When an RVC4 DAI benchmark ran with device monitoring enabled and no explicit device was selected, DepthAI and the monitor could independently choose devices. This could make the benchmark run on one RVC4 while monitoring another.

This change resolves the automatic RVC4 target once for monitored DAI runs, then passes the same resolved target to both the DepthAI benchmark and the device monitor.

Explicit device selection is preserved, unmonitored DAI keeps DepthAI auto-selection, and SNPE benchmarking is unchanged.

## Tests

- Added regression tests for monitored DAI same-device selection and unmonitored DAI auto-selection
- Targeted unit tests pass
- Repository `prek` hooks pass
- Verified on hardware that the resolved DepthAI and monitor targets match

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved automatic device selection for monitored benchmarks when no device address is configured.
  - Ensured benchmarking and device monitoring consistently use the same connected device.
  - Preserved automatic device selection for benchmarks that do not use monitoring, avoiding unnecessary device targeting.
  - Added clearer handling when no compatible device is available.


modelconverter/platforms/rvc4/benchmark.py
```diff
@@ -88,6 +91,24 @@ class InputSpec(BaseModelExtraForbid):
 }
 
 
+def _get_first_rvc4_device_info() -> dai.DeviceInfo:
```
klemen1999, line 94: Nitpick: Let's put these functions below class definition

klemen1999 (approved):
Generally LGTM

modelconverter/platforms/rvc4/benchmark.py
rolandocortez: Addressed in [1fab8ee](https://github.com/luxonis/modelconverter/pull/312/commits/1fab8ee0e63a2f992af3b8dc487ae89b2762eb94)

